Issue #1, opened 2018-01-03
Update to Checkstyle 8.7


lkoe:
Update the plugin to Checkstyle 8.7

lkoe:
#2

romani:
Please review labels at checkstyle project, if you agree with them, we can start to use them in this project too. All of them targeted to let contributor known status of issue, the most critical is label "approved" to make it clear that fix is wellcome we all agree on fix. Issues without it are in discussion or incomplete or ... .

romani:
Additional point is that 8.7 release is special release to just introduce painful but easy api damage.
It would be awesome to make release to 8.6 , as only it contains metadata changes and next do one more release that will demand from you only version bump

lkoe:
Regarding labels: yes, good idea to align this approach. 

Regarding separate 8.6/8.7 releases: I'd rather not since it's quite some work for me and the intermediary release holds no value für plugin users. I agree, however, that having releases in "lock-step" would be preferable in the future.

romani:
@lkoe , I created milestone for you https://github.com/checkstyle/eclipse-cs/milestones
it is better to mark issue with milestone to share where/when it will be released, to avoid extra conversations from comunity.

rwalkerands:
Is this release now meant to be available on the update site? (It still points to 8.5.1.)

romani:
It is implemented but not released yet, https://github.com/checkstyle/eclipse-cs/releases 

@lkoe , it would be awesome if you just make a release. Do not wait for extra fixes, just checkstyle update is more than enough. Sevntu release is blocked by eclipse-cs .

lkoe:
@romani I'll do it asap, but likely not before Jan 13

PR #2, opened 2018-01-03
Issue #1: update to Checkstyle 8.7

lkoe:
Issue #1: update to Checkstyle 8.7


romani:
I is hard to review now, what I recommend is to reference issue in commit message and PR description.
And have only one commit.
I recommend eventually come up to our rules for PRs https://github.com/checkstyle/checkstyle/wiki/PR-rules ,let me know if you agree and when you would like to follow.

I also changed settings of project to allow only "rebase and merge" commits as it helps a lot in git history investigations.

lkoe:
Thanks Roman, I was still toying around with the tools. Also my commit user/email was off. Should be better now.

romani:
In future it is better to not mix different changes in one commit.
it is better to have several commits or several issues(PRs), it helps others project maintainers to understand what forced you to make changes like this.

I know that is a bit more complicated(bureaucracy), but as we work together we should share a bit more context of why changes are required.

lkoe:
> In future it is better to not mix different changes in one commit.

agreed :-)

PR #13, opened 2018-01-11
fix Maven warnings

Bananeweizen:
The tycho-source-plugin version was overridden in the children of the
aggregator instead of inheriting the version. This was necessary because
the parent version definition had a typo (tycho.version instead of
tycho-version). After fixing the typo, the overrides are not necessary
anymore.


lkoe:
Great catch! This was bothering me quite some time...

Bananeweizen:
I was similarly confused because I deleted the overrides first and then suddenly had Maven errors for something that worked in many other projects before. :)

PR #16, opened 2018-01-12
fix #11: have a target definition

### LINKED ISSUE #11: Missing target platform configuration
The repository does not contain a .target file to configure the target platform for the plugin development environment. That way, the plugin sources are always compiled against the bundles of the running Eclipse IDE, thereby easily introducing incompatible changes for older versions of Eclipse.

Did I miss something here or shall I create and contribute such a target definition?

Bananeweizen:
This introduces an explicit target definition to avoid incompatible
changes. Developers need to open the .target file and hit "Set as active
target platform". The Tycho build uses the target automatically.

Java changes are just to get back to the old style adapter
implementation which requires an explicit cast. Manifest change is to
remove a seemingly unused e4 dependency.

I'm running an installation that uses the binary produced by this change
and it works fine.


Bananeweizen:
Is there any way to easily retrigger a travis build as non project member? I'm quite sure the error was only caused by some network connection trouble (a plugin not being downloaded from some eclipse mirror) and will vanish when being retriggered.

net.sf.eclipsecs.ui/META-INF/MANIFEST.MF
```diff
@@ -23,7 +23,6 @@ Import-Package: org.eclipse.core.commands,
  org.eclipse.core.runtime,
  org.eclipse.core.runtime.jobs,
  org.eclipse.core.runtime.preferences,
- org.eclipse.e4.ui.model.application;resolution:=optional,
```
lkoe: I think this dependency is still needed, as a few Eclipse3/4 switches depend on it, see https://github.com/checkstyle/eclipse-cs/blob/7eb81a2068e12ec0d9accee30db2ea8dbec3de49/net.sf.eclipsecs.ui/src/net/sf/eclipsecs/ui/CheckstyleUIPlugin.java#L135

This could be refactored of course if Eclipse 3 compatibility is dropped altogether.

Bananeweizen: Ah, good advice. In that case, Indigo is not sufficient. I had taken that since the feature.xml required at least eclipse.core 3.7. I will check the class name references more closely and come up with a modified PR that includes a slightly higher Eclipse version as target. Thanks.

PR #19, opened 2018-01-13
Issue #18 project specific formatter settings

lkoe:
Issue #18 Adds project specific formatter settings

PR #20, opened 2018-01-13
configure m2e on the new target project

Bananeweizen:
When I created the target project, I forgot to configure it for the
Maven Eclipse integration m2e (like all the other projects).


lkoe:
I think the same changes are also in my PR https://github.com/checkstyle/eclipse-cs/pull/19

Bananeweizen:
Right, therefore closing here.

PR #27, opened 2018-01-15
Prepare website for Github pages

lkoe:
Generates the eclipse-cs website into the <repository-root>/docs folder.
From there it can be served by Github pages, see examplary results here: https://lkoe.github.io/eclipse-cs

Bonus point: Website can simply be published during the "prepare and build release" step


lkoe:
>Output looks good to me. If I get it right, this is mostly a move of the existing directory and some modification of the POM to produce it? Besides a coarse scan, I cannot judge this any further, since I'm not into web development that much.

Yes, that's about the gist of it. Thanks for having a look.
The site is now published: https://checkstyle.github.io/eclipse-cs

I'll put a redirect on the sourceforge page and update the release instructions.

romani:
@lkoe , @Bananeweizen , you are affecting all custom Checks projects, details - https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/639 .

It might be reasonable to to keep dtd files on sourceforge for some time if possible, or make big announce in news , about this movement to new website.

lkoe:
@romani sorry, this wasn't intentional. I've re-uploaded the dtds to sourceforge for the time being.

PR #29, opened 2018-01-16
advertise additional checks

Bananeweizen:
Have a reference to the update site of the additional checks in our
update site. That way it is automatically added to the list of known
update sites in the local P2 update manager. However, it is initially
disabled, and must be enabled manually by the user (so no additional
checks get installed by this change).

Also change the update site URL inside the update site description
(there is actually no field in the form based category editor showing
this property).

### DIFF
--- a/net.sf.eclipsecs-updatesite/category.xml
+++ b/net.sf.eclipsecs-updatesite/category.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <site>
-   <description name="Eclipse Checkstyle Plugin" url="http://eclipse-cs.sourceforge.net/update">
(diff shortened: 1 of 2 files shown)


lkoe:
Hmm, I am not so sure about this one. 
Sevntu is already added on the online update site (see https://github.com/checkstyle/eclipse-cs/blob/master/net.sf.eclipsecs.doc/src/main/resources/update/compositeArtifacts.xml).
If I understand correctly this change would also add it to the packaged update site binary (?) - which up until now is self-contained.

Bananeweizen:
Oh, I completely missed the inclusion in the update site, although I have seen it before. I'm okay with abandoning this change then...

PR #30, opened 2018-01-16
unify installation instructions

Bananeweizen:
Have all versions from Juno to Oxygen in all places mentioning
installation. As an alternative, we could simplify this to "...into a
running Eclipse (Juno or newer version) workspace..."


lkoe:
> As an alternative, we could simplify this to "...into a
running Eclipse (Juno or newer version) workspace..."

Good point, we should do that for the next release. 👍

PR #33, opened 2018-01-17
Deploy update site and update site archive to bintray using "mvn deploy"

lkoe:
Deploys the update site and update-site.zip to bintray.

Add the following to your .m2/settings.xml:
```
	<servers>
		<server>
			<id>bintray</id>
			<username>yourBintrayUser</username>
			<password>yourBintrayApiKey</password>
		</server>
	</servers>
```

Then it's a matter of running `mvn deploy -P bintray` to build and publish the binaries.

@Bananeweizen I sent you an invite for the eclipse-cs organisation on Bintray yesterday, please check if you've received that
@romani I couldn't sent an invite yet, because you don't seem to have an account there

### DIFF
--- a/net.sf.eclipsecs-updatesite/pom.xml
+++ b/net.sf.eclipsecs-updatesite/pom.xml
@@ -9,4 +9,76 @@
     <artifactId>net.sf.eclipsecs-updatesite</artifactId>
     <packaging>eclipse-repository</packaging>
     <name>eclipse-cs Update Site</name>
+
+    <profiles>
+        <profile>
+            <id>bintray</id>
+            <build>
+                <plugins>
+                    <plugin>
+                        <artifactId>maven-antrun-plugin</artifactId>
+                        <executions>
+                            <execution>
+                                <phase>deploy</phase>
+                                <configuration>
+                                    <target>
+                                        <copy todir="${project.build.directory}/archive/">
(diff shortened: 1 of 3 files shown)


lkoe:
Caveat: the bintray-maven-plugin version used does not support proxy authentication, yet. 

When https://github.com/random-maven/bintray-maven-plugin/issues/1 is available as official version, the plugin version used should be upgraded.

Bananeweizen:
Okay, I've setup my settings.xml and joined the organization. How does the POM deal with uploading something under the exact same version a second time? Since you mentioned having corrupted the existing 8.7.0 in another issue, can we configure the upload to abort if the version already exists on bintray? I think I would rather like it fault-safe, and in case of actually wanting to overwrite an existing version, I'd be happy to modify the pom at that time.

lkoe:
> How does the POM deal with uploading something under the exact same version a second time?

It doesn't. It will simple upload the files it is being pointed at. Problem is the build timestamp changing everytime. Previously I was using a different timestamp provider based on the latest commit, which - again theoretically - provides stable build timestamps. Problem with that was that some sub-projects ended up with a different timestamp than others for reasons I could not figure out.

In any case it is not required to use the deploy target. You can still do it manually, but it's a bit tiresome.

Issue #32, opened 2018-01-16
Adapt issue markers to show issue type


Bananeweizen:
When using "Properties" on an issue in the problems view, that shows just the properties available from the marker framework. I would like to try adapting the marker in such a way, that it also shows the related checkstyle style as additional property in the same dialog.

This might help in more easily locating the check in the configuration pages, since just from the issue text that is not always obvious.

Bananeweizen:
Was already implemented.

PR #35, opened 2018-01-17
#32: new checkstyle marker property page

Bananeweizen:
This adds a new property page for checkstyle issues, which show the
check and module of the issue.

This is incomplete, since the retrieved module and key need to be
localized.

### DIFF
--- a/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/config/Module.java
+++ b/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/config/Module.java
@@ -30,7 +30,7 @@ import net.sf.eclipsecs.core.config.meta.MetadataFactory;
 import net.sf.eclipsecs.core.config.meta.RuleMetadata;
 
 /**
- * Object representing a module from a checkstyle configuration. Can be augumented with meta data.
+ * Object representing a module from a checkstyle configuration. Can be augmented with meta data.
  * 
  * @author Lars Ködderitzsch
  */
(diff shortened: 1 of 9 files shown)


Bananeweizen:
2018-01-17_18h53_29
The change looks like this, when opening properties on an issue from the problems view. The labels in red need localization. @lkoe Do you have a hint how to get the localized version of those, e.g. similar like in the configuration dialogs somehow?

lkoe:
> @lkoe Do you have a hint how to get the localized version of those, e.g. similar like in the configuration dialogs somehow?

For localizing the module name you can traverse the [Metadata](https://github.com/checkstyle/eclipse-cs/blob/8474717f6b379b30b909facd8fdab35027cb5a3f/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/config/meta/MetadataFactory.java#L143).
Regarding the message key itself (`name.invalidPattern`), what did you intend to display there? The raw message pattern?

Bananeweizen:
I think I confused the invalidPattern, since that is the same as the error message in the first line, right? Basically I want to display these properties from the configuration screen, such that users are able to locate the necessary configuration more easily after getting a problem marker:

wanted_properties

lkoe:
Ok, that would be group name, module name and description, all accessible from https://github.com/checkstyle/eclipse-cs/blob/8474717f6b379b30b909facd8fdab35027cb5a3f/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/config/meta/RuleMetadata.java#L33

Bananeweizen:
Thanks, that was very helpful. I consider this finished now. Looks like this when hitting Ctrl-Enter on a problem marker in the problems view:
property_page_finished

I was wondering if the description browser should also be set to take the background color. But if so, then this should be applied to all configuration dialogs as well as this property page (in a separate change).

lkoe:
This looks pretty useful. Personally, I am ok with the background color - but if you feel it would look better with the same background, go ahead.
Only thing I would consider aesthetically was maximizing the available space for the description by wrapping it beneath the "Description:" label and give it the full available width. But again this is a matter of personal taste, which can be argued about :-)
I would also accept this as is.

Bananeweizen:
Oh, that is a good proposal, I will rework it. Such layout changes can improve usability a lot and tend to be ignored by developer too often, since we have our head to deep into the algorithm parts.

PR #41, opened 2018-02-04
release version 8.8

Bananeweizen:
This is according to the first 6 steps of the release wiki page. I have not attempted updating any binaries or listings.


Bananeweizen:
Who should do that? I think I can upload to bintray, since you added me to the group there, but I guess I cannot release on sourceforge, github and marketplace. Therefore I assume you do all that together?

lkoe:
Release done, sorry for the long delay.

PR #51, opened 2018-03-05
fix or silence generics warnings

Bananeweizen:
First I fixed all warnings that could be fixed by adding generic type
arguments. Only afterward I silenced the remaining warnings (basically
in all the AST visit() methods which operate with the non-generic JDT
API). Additionally, all adapter related assignments have been silenced, since we still compile against the non-generic adapter implementation.

### DIFF
--- a/net.sf.eclipsecs.ui/src/net/sf/eclipsecs/ui/stats/views/AbstractStatsView.java
+++ b/net.sf.eclipsecs.ui/src/net/sf/eclipsecs/ui/stats/views/AbstractStatsView.java
@@ -36,6 +36,7 @@ import net.sf.eclipsecs.ui.stats.views.internal.CheckstyleMarkerFilterDialog;
 
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IMarkerDelta;
+import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IResource;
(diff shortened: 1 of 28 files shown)


net.sf.eclipsecs.ui/src/net/sf/eclipsecs/ui/stats/views/AbstractStatsView.java
```diff
@@ -296,7 +298,7 @@ protected final IDialogSettings getDialogSettings() {
    */
   private void focusSelectionChanged(IWorkbenchPart part, ISelection selection) {
 
-    List resources = new ArrayList();
+    List<IResource> resources = new ArrayList<IResource>();
```
vorburger: new ArrayList<>(); - but never mind... ;-)

vorburger:
+1 LGTM

net.sf.eclipsecs.ui/src/net/sf/eclipsecs/ui/stats/views/AbstractStatsView.java
Bananeweizen: I'm still not sure why Eclipse does this sometimes when adding generic arguments. Must be my fault. Besides that, I still want to run some cleanup actions on the current code, where it should be more easy to remove this again, so I would leave it to that.

PR #52, opened 2018-03-07
refactor: replace deprecated instance scope

Bananeweizen:
The new API also works with Eclipse 3.x.


vorburger:
I've no idea what exactly `InstanceScope` is, but I've looked this proposed change over, and I guess there is now a static `INSTANCE` singleton, so seeing that it does still pass the build, and knowing that @Bananeweizen typically knows what he is doing, I will say _LGTM +1_ here and recommend that whoever has merge permissions on tihs repo accept this proposed change.

Bananeweizen:
@vorburger Nice to see you here, Michael. I've only recently become a co-maintainer of this plugin, so I was still a bit hesitant, but I can actually merge it myself. :)

PR #53, opened 2018-03-10
fix multiple extension declaration issues

Bananeweizen:
* remove empty [...] sections, they are otherwise part of the generated
documentation
* replace link tags by code tags, since link tags lead to validation
errors and are not rendered
* use descriptive names for the extension points instead of repeating
their IDs
* fix the wrong occurrences of "filters" as extension point schema IDs.
* typos

None of these changes influences the extension point registry. All
relevant IDs are still the same as before, so extenders continue to work
without modification.

### DIFF
--- a/net.sf.eclipsecs.core/plugin.xml
+++ b/net.sf.eclipsecs.core/plugin.xml
@@ -7,7 +7,7 @@
         schema="schema/checkstyleAddonProvider.exsd"/>
 
     <!-- defines the extension point to contribute custom configuration type. -->
-    <extension-point id="configurationtypes" name="Configuration types" schema="schema/configtypes.exsd"/>
+    <extension-point id="configurationtypes" name="Checkstyle configuration types" schema="schema/configtypes.exsd"/>
 
     <!-- defines the extension point to contribute default checkstyle configurations to the checkstyle plugin -->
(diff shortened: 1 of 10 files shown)


lkoe:
Hi, it seems you're committing (at least) XML-Files with CRLF line separators (Windows). That's why the whole file appears as changed and no proper diff.
Please check you git client settings, especially `core.autocrlf`, which should be `true` on Windows, see also https://help.github.com/articles/dealing-with-line-endings/

Bananeweizen:
Nope, this is not a git issue per se. The project properties for line delimiters are set to "inherit from container", which makes them "Windows" when running Eclipse on Windows (and the text based Eclipse editors just take that project property). I'll check-in modified .prefs to enforce unix line feeds, if that is what the majority of files has right now. That way, at least when editing in Eclipse, this should be correct independent of the git settings.

lkoe:
If that works for you, ok. I am not so sure it will affect existing files, though. In any case, please fix the line endings in this PR.

The git client setting would automatically transform all CRLF to LF upon commit (and vice versa on checkout), so you could very well work with Windows separators locally.

Bananeweizen:
I have converted the line endings. Just as a note for review: All changes have been done using the extension schema editor (instead of plain text editor), so deleted sections or deleted namespaces should all be fine.

PR #59, opened 2018-03-19
Fix typo

Bananeweizen:
One of the occurrences is a label in a dialog, found it that way. All
others are just Java code or comments.


net.sf.eclipsecs.ui/src/net/sf/eclipsecs/ui/properties/filter/PackageFilterEditor.java
```diff
@@ -420,23 +420,23 @@ public void setSize(int width, int height) {
...
     /**
-     * Sets if subtree should be recursivly excluded. Default is true.
+     * Sets if subtree should be recursively excluded. Default is true.
      *
-     * @param recursivlyExcludeSubTree
+     * @param recursivelyExcludeSubTree
      *          the recursive checking state
      */
-    public void setRecursivlyExcludeSubTree(boolean recursivlyExcludeSubTree) {
+    public void setRecursivelyExcludeSubTree(boolean recursivelyExcludeSubTree) {
```
vorburger: This (and next) of course technically break a public API, but I'm sure it's not a problem here.

Issue #70, opened 2018-04-30
upgrade to checkstyle 8.10


romani:
http://checkstyle.sourceforge.net/releasenotes.html#Release_8.10

romani:
is there any plans to update version of checkstyle in eclipse-cs and make a release ?

if no updates except for checkstyle version bump, it is still a reason to do release. Users use checkstyle in multiple build systems (Eclispe-cs -> maven -> sonar ) so to reliably upgrade to new version they have to upgrade version of checkstyle in all plugins.

issue is detected base on conversation at https://github.com/OpenGamma/Strata/pull/1688

rtack:
looks like this could be closed see https://github.com/checkstyle/eclipse-cs/releases/tag/8.10.0

PR #72, opened 2018-05-12
#70: upgrade to checkstyle core 8.10

Bananeweizen:
* upgrade core library
* create release notes
* upgrade version numbers
* moved alert message of Lars out of each version, now on top of release
notes
* typo fixes in metadata
* packaged and updated documentation

There haven't been any changes in checkstyle/xdocs mentioning a newly
introduced config element since 8.6, so it seems safe to assume there
are no metadata changes.

### DIFF
--- a/docs/partials/builtin-config.html
+++ b/docs/partials/builtin-config.html
@@ -14,8 +14,7 @@
     <p>
         <b>Example:</b>
     </p>
-    <pre><code>
-&lt;extension
+    <pre><code>&lt;extension
     id="checkstyle.CheckConfiguration"
     point="net.sf.eclipsecs.core.configurations"&gt;
     &lt;check-configuration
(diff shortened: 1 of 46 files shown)


romani:
Not a big deal for 8.8 -> 8.10 but ...

Can we release eclispe-cs for each version of checkstyle ? even it will be two releases in the same date. It will help users to upgrade to required version and be consistent by checkstyle version in all environments (maven, eclipse, idea, sonar, ...).
Possible use case: user got released fix for problem in 8.9, but in 8.10 we did breaking compatibility that affects user or unfortunate regression or ..... , so user can not upgrade to 8.10, but still wants fixes of 8.9.  

We follow this model in sonar: https://github.com/checkstyle/sonar-checkstyle/blob/master/README.md
in idea - https://github.com/jshiell/checkstyle-idea#version-history , releases on each fix and have ability to use any version of checkstyle.

Bananeweizen:
This depends very much on the availability of @lkoe. The code already had the update to 8.9 since 5 weeks, but it never got released.

That being said, Eclipse users are generally not used to choosing plugin versions. Either they are working with what they have at the moment, or they upgrade to the latest available. Also the complete UI and update procedure target the use case of either updating or not updating. But no selection in between.

And to my opinion, the problem of breaking configuration files must be solved by the checkstyle core team. Honestly, I really don't understand why your XML structure nesting must replicate the subclassing of the checks, to me that has absolutely no benefit, and is a violation of the separation of concerns principle. In the XML file, make every check a direct child of the root node, and that's it. No more refactoring of config files, when moving stuff into the treewalker or vice versa. It should not matter, when a check A is a child of class B is a child of class C. For configuration purposes, it is only relevant that it is called "A", so just put the A config directly under the root node.

Bananeweizen:
Looks good. I updated version number, and both support url and homepage URL (so they point to github issues and github.io now).

PR #79, opened 2018-05-19
Fix #3 Automated metadata testing + metadata corrections (bulk of work by @rnveach)

### LINKED ISSUE #3: Automate metadata verification using reflection
As discussed at https://sourceforge.net/p/eclipse-cs/discussion/274376/thread/3b44a5eb/# , we should implement sevntu's metadata verification to make updates to the metadata easier.
This will ensure updates aren't missed and that the meta file is defined correctly.

lkoe:
Adds a unit test to the `net.sf.eclipsecs.checkstyle` project which uses reflection to check eclipse-cs metadata for completeness.
Largely based on a contribution from @rnveach


romani:
@lkoe and @Bananeweizen , looks like huge amount of work stuck in review.
I might be wrong but, points are kind of minor and could be resolved in separate issue. It is good to keep them noted somewhere , but pass huge amount of metadata fixes.

if that is possible to speedup merge on this PR ?

lkoe:
I'll have a look at this today. Sorry for letting it slide so long.

lkoe:
@Bananeweizen thanks for the review. I've now removed all unrelated project files and reimplemented the 3 suggested methods.

PR #88, opened 2018-07-19
Fix #84: upgrade to checkstyle 8.11

### LINKED ISSUE #84: upgrade to checkstyle 8.11
http://checkstyle.sourceforge.net/releasenotes.html#Release_8.11

valfirst:


romani:
@lkoe , @vorburger , it might be reasonable to release update to 8.10.1 first, as it has bunch of metadata updates.

Unfortunately 8.11 have some sort of breaking changes (renaming of Utils classes) that might affect all possible third-party extensions with Checks (just one example: https://github.com/sevntu-checkstyle/sevntu.checkstyle). 
So such extension will must to be recompiled!!!
It will take additional time for users to migrate, so it is better to make keep in separate releases big fix for metadata (#3) and upgrade to 8.11 .

valfirst:
#89: upgrade to checkstyle 8.10.1

lkoe:
Thanks, after merging your first PR this needed to be rebased, see https://github.com/checkstyle/eclipse-cs/pull/90

PR #108, opened 2018-10-23
Add filter for Files Older Than One Day

pheaber:
Can be useful for massive legacy projects when the "in sync with repo" doesn't work. This way as you modify files, Checkstyle is executed against them.

### DIFF
--- a/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/projectconfig/filters/FilesOlderThanOneDayFilter.java
+++ b/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/projectconfig/filters/FilesOlderThanOneDayFilter.java
@@ -1,0 +1,25 @@
+package net.sf.eclipsecs.core.projectconfig.filters;
+
+import org.eclipse.core.resources.IFile;
+
+public class FilesOlderThanOneDayFilter extends AbstractFilter {
+
+	private static final long MILLIS_IN_24_HOURS = 1000 * 60 * 60 * 24;
+
+	@Override
+	public boolean accept(Object o) {
+		boolean goesThrough = true;
+
+		if (o instanceof IFile) {
+			IFile file = (IFile) o;
+			if ((System.currentTimeMillis() - file.getLocalTimeStamp()) < MILLIS_IN_24_HOURS) {
+				goesThrough = true;
+			} else {
+				goesThrough = false;
+			}
+		}
+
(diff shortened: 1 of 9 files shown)


romani:
Please make reference from commit message to PR (Pull #108).

Please explain in description of PR new feature in detail.

pheaber:
This adds another option to the "Exclude from checking" section of the Checkstyle preferences. When selected, it will exclude any files that are older than one day. 

We've found it useful since we have a really large legacy code base that we only want our Checkstyle rules executed against files as they are changed. The existing "in sync with repo" option didn't work for us, originally since we used an odd source control but even now on git it seems to not run against some modified files.

I'm not sure what you mean for me to reference the PR from the commit.

rnveach:
@pheaber This branch needs to be rebased.

net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/projectconfig/filters/FilesOlderThanOneDayFilter.java
```diff
@@ -0,0 +1,21 @@
...
+	@Override
+	public boolean accept(Object o) {
+
+		if (o instanceof IFile) {
+			IFile file = (IFile) o;
+			if ((System.currentTimeMillis() - file.getLocalTimeStamp()) < MILLIS_IN_24_HOURS) {
+				return true;
+			}
+		}
+		return false;
```
Bananeweizen: All other filter implementations let unknown stuff pass (i.e. they return true by default). Your filter should behave the same and only return false at the end of the instanceof block. Otherwise if there are non-file objects, your filter always rejects them.

romani:
@pheaber , please update commit message to reference PR number like `Pull #108: ...<your text>.` , it will help us in future tor track change up to this PR where some ideas and details are shared.

there are two approvals ... so your PR is ok to merge.

pheaber:
Oh, my bad. Updated commit message as @romani requested. Thanks for explaining that better for me, I didn't realize you could edit the commit message.

rnveach:
@pheaber there are still conflicts.
Instead of us merging 6+ commits, can you rebase off latest master and squash all your commits to 1?

pheaber:
Hopefully I merged those correctly

Bananeweizen:
I've submitted the change as new PR, since this one is completely broken.

PR #109, opened 2018-10-24
allow parallel building with checkstyle

Bananeweizen:
Since Eclipse Photon or Oxygen users can enable parallel building of the
workspace. However, parallel building requires all existing project
builders to implement a useful scheduling rule. If builders don't
implement that method, they automatically lock the workspace root.

A simple and still safe implementation for checkstyle is to lock only
the currently analyzed project. That allows all other builders to run on
the remaining projects in parallel.

See https://de.slideshare.net/mickaelistria/parallel-builds-in-eclipse-ide-workspace
for more technical background.

### DIFF
--- a/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/builder/CheckstyleBuilder.java
+++ b/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/builder/CheckstyleBuilder.java
@@ -54,6 +54,7 @@ import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.OperationCanceledException;
 import org.eclipse.core.runtime.Status;
+import org.eclipse.core.runtime.jobs.ISchedulingRule;
 import org.eclipse.osgi.util.NLS;
 
 /**
@@ -206,7 +207,7 @@ public class CheckstyleBuilder extends IncrementalProjectBuilder {
    * Builds the selected resources.
    *
    * @param resources
-   *          the resourcesto build
+   *          the resources to build
    * @param configuration
    *          the project configuration
    * @param monitor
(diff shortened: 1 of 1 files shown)


romani:
@Bananeweizen , it is better to wait for approval from person of Checkstyle organization.
all approvals from outside are not qualified.

Bananeweizen:
@romani I don't get that sentence. melloware is a committer, what different people would you expect? At least for the eclipse plugin project, I'm not aware of doing a "steward" review model, where only a privileged group of committers can approve changes. If another committer reviews, that is sufficient to me.

melloware:
@Bananeweizen I am not a committer on this project.  I am a big fan of it and your change to me seems like an excellent one worth committing.  But yes I don't see any "steward" review model posted.

rnveach:
As far as I know it is the same for all github projects. Only admins of the project can merge stuff, which is only people in the `checkstyle` organization here. Even if me or @romani approved this, we would probably still want someone from eclipse-cs ( @Ikoe ) to look it over as we aren't plugin maintainers and we don't know how this change would affect the plugin in eclipse.

Bananeweizen:
Thanks for the comments. In that case I may have been fooling myself. I thought I already had one other eclipse based review now, and that the merge button was enabled because of that. But probably that was just because I'm a committer on the eclipse plugin repository myself.

Nevertheless, the same change has been applied to the Eclipse Spotbugs, JDT, PDE, and m2e plugins already, so it should be working fine. :)

(Note for testers: If you want to check if performance is improved after this change, then all active implementations of IncrementalProjectBuilder must return a non workspace root scheduling rule. If just a single builder returns the workspace root, then _all_ builders will build with the workspace root as scheduling rule. Therefore testing must happen in a heavily patched eclipse installation as of now.)

romani:
>If another committer reviews, that is sufficient to me.

It is better to stay to approve merge by people who have RW access to repository (maintainers). If any disagreement happened between maintainers, there should be a pause and disagrement should be resolved by any means.
Pleople outside maintainers team are always welcome to share ideas and do review, but they do not allow merge. Only maintainers know what needs to be fixed and how to test and .... .

PR #110, opened 2018-11-01
Hint with rule name for suppression

Bananeweizen:
In the marker properties dialog have a hint showing the internal rule
name. While the human readable name is useful to locate a check in the
configuration dialog, the internal rule name is needed to create
suppression comments.


rnveach:
@Bananeweizen ping

romani:
@Bananeweizen, ping.
Let's finish this PR in any form, some step forward is better than no step at all

romani:
moved to #173 to merge as is.

PR #112, opened 2018-11-01
Fix #111: Add filter for files older than one day

### LINKED ISSUE #111: Add filter for Files Older Than One Day
Can be useful for massive legacy projects when the "in sync with repo" doesn't work. This way as you modify files, Checkstyle is executed against them.

Taken from https://github.com/checkstyle/eclipse-cs/pull/108

Bananeweizen:
Issue https://github.com/checkstyle/eclipse-cs/issues/111

Cherry picked and cleaned up from https://github.com/checkstyle/eclipse-cs/pull/108

### DIFF
--- a/net.sf.eclipsecs.core/core.properties
+++ b/net.sf.eclipsecs.core/core.properties
@@ -34,3 +34,6 @@ WriteProtectedFilter.label       = write protected files
 
 DerivedFilesFilter.description   = Filters files marked as derived, which often applies to generated sources.
 DerivedFilesFilter.label         = derived (generated) files
+
+FilesOlderThanOneDayFilter.description  = Exclude all files that are older than one day.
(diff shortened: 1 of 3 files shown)


rnveach:
Original changes seen at https://github.com/checkstyle/eclipse-cs/pull/108/commits/55a3bd196ee61c517cb133626eb1d28cb700d721 minus message.properties

Bananeweizen:
@rnveach That's why I did this. message.properties was never part of the filter change, but it was merged a while ago with https://github.com/checkstyle/eclipse-cs/commit/7009a3867a8309c072747df8fcb9c94e8c0eb655 already. Therefore 3 files only should be correct.

rnveach:
I only added the comment to compare what was already approved to what is now in a new PR to make sure nothing was lost. I don't have the very original commit to compare to.

PR #118, opened 2018-12-27
update Maven plugins

Bananeweizen:
Update all the Maven plugins which have a fixed version. To simplify the
check for Maven updates in the future, a new shared launch config for
Eclipse has been added.

Still, some standard Maven plugins (and Maven itself) are not versioned,
which can lead to different build results, and should be changed.

### DIFF
--- a/net.sf.eclipsecs.target/Maven update check.launch
+++ b/net.sf.eclipsecs.target/Maven update check.launch
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<launchConfiguration type="org.eclipse.m2e.Maven2LaunchConfigurationType">
+<booleanAttribute key="M2_DEBUG_OUTPUT" value="false"/>
+<stringAttribute key="M2_GOALS" value="versions:display-dependency-updates versions:display-plugin-updates"/>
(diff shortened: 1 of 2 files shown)


romani:
I never used maven version updates from Eclipse UI.

But why not simply use maven standard plugin for controlling this example: https://github.com/checkstyle/checkstyle/blob/master/config/version-number-rules.xml 
We run it as: https://github.com/checkstyle/checkstyle/blob/master/.ci/travis/travis.sh#L97
Result: https://travis-ci.org/checkstyle/checkstyle/jobs/472422393#L2843

Bananeweizen:
> But why not simply use maven standard plugin for controlling this ?

The new launch config does nothing more than running "versions:dependency-updates versions:plugin-updates" from eclipse, so it is exactly the same as you running this from a wrapper shell script. The launch config is only meant to help developers that they don't need to remember the syntax (I tend to forget it, too).

> example: https://github.com/checkstyle/checkstyle/blob/master/config/version-number-rules.xml

Version plugin rules are deprecated. See the plugin documentation.

romani:
It will stay :) as it awesome functionality, it is marked as deprecated by mistake.

https://www.mojohaus.org/versions-maven-plugin/version-rules.html
https://github.com/mojohaus/versions-maven-plugin/issues/157#issuecomment-403890568

PR #121, opened 2018-12-29
Fix #120: checkstyle version in preferences

### LINKED ISSUE #120: Show the checkstyle version in eclipse preferences
It would be useful to show the current running version of checkstyle somewhere on the preferences page, for informational purposes.

Bananeweizen:
Show the version as the preference page "description" field. The displayed version is the one of the embedded JAR, independent of the version of the Eclipse plugin (e.g. for the current plugin version 8.12 it shows 8.13 correctly).

2018-12-29_15h50_05

### DIFF
--- a/net.sf.eclipsecs.ui/src/net/sf/eclipsecs/ui/Messages.java
+++ b/net.sf.eclipsecs.ui/src/net/sf/eclipsecs/ui/Messages.java
@@ -174,6 +174,8 @@ public final class Messages extends NLS {
 
   public static String CheckstylePreferencePage_txtSuggestRebuild;
 
+  public static String CheckstylePreferencePage_version;
+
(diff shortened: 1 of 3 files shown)


romani:
In case you can use smth like https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/Main.java#L153

Bananeweizen:
> In case you can use smth like https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/Main.java#L153

Great! I had seen the related entry in the manifest of checkstyle when looking for available version numbers, but until now I did not know this mechanism of packages to get a vendor and version. Thanks a lot.

Bananeweizen:
> @Bananeweizen squash all commits into 1 please.

Done.

PR #124, opened 2018-12-30
Fix some checkstyle violations

Bananeweizen:
This fixes the checkstyle violations that only have a small number of
occurrences:
* wrong Javadoc see tag
* not commented throws javadoc tags
* tabs versus space

### DIFF
--- a/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/builder/Auditor.java
+++ b/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/builder/Auditor.java
@@ -416,10 +416,8 @@ public class Auditor {
      *          the audit error
      * @param markerAttributes
      *          the marker attributes
-     * @throws CoreException
      */
-    private void calculateMarkerOffset(AuditEvent error, Map<String, Object> markerAttributes)
-            throws CoreException {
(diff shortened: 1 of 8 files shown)


rnveach:
@Bananeweizen @lkoe If you are going to fix checkstyle violations in the project, why don't you add checkstyle to the verify process of your project so the CI can always validate it on a PR.
It is pointless to manually fix things, unless you enforce that they can't come back.

romani:
Ok to merge, but highly recommened to endorse such violations to never come back

Bananeweizen:
I must admit I'm not aware how to configure checkstyle to compare new to old results, except when using the Jenkins static analysis plugin. But even there it only compares previous build against current build, which is not what we want (we want current commit results against previous commit results). Therefore someone else would need to configure that.

rnveach:
I wouldn't do a comparison of checkstyle results between old and new. I would enforce 0 violations on the project. You can use suppressions to hide violations from code not yet fixed and go through either files or checks, one by one, enabling them and removing the suppressions as you go.

romani:
Just most recent example when we postponed resolving some violations to later 89c8642bc51cc22bcab887f4868dacf799a9dbb1

PR #155, opened 2019-05-03
fix typos and whitespace in metadata

Bananeweizen:
Several metadata translations end in superfluous horizontal tab
characters (encoded as u0009). That leads to mouse hovers showing
additional "free space" after the text. Remove all those whitespace line
endings and also fix all typos found during the cleanup.

image


Bananeweizen:
rebased only, will merge after travis build

PR #173, opened 2019-10-31
Hint with rule name for suppression

romani:
from #110

In the marker properties dialog have a hint showing the internal rule
name. While the human readable name is useful to locate a check in the
configuration dialog, the internal rule name is needed to create
suppression comments.


romani:
merged without testing of it.

PR #178, opened 2019-11-06
Fix releases json format

cknoblauch:
A couple of trailing commas are breaking the rendering of the changelog page.


romani:
thanks a lot, I merged manually with commit message amendments.

Issue #163, opened 2019-06-23
upgrade to checkstyle 8.22


romani:
https://checkstyle.org/releasenotes.html#Release_8.22

Dichotomia:
What is the blocking point for a new release ?

lkoe:
Time and motivation

Dichotomia:
A motivation could be the possibility to use new checks introduced by this version like "OrderedProperties".

For example, i had to uninstall this plugin, because the code verification fail in Eclipse and the plugin crash. The cause was the activation of new checks.

romani:
@Dichotomia , please be welcome to contribute, git history has bunch of examples on what is required to update to make it happen, we will guide you through the process.

rnveach:
Upgrade should start with earlier versions first.

Dichotomia:
> 
> 
> @Dichotomia , please be welcome to contribute, git history has bunch of examples on what is required to update to make it happen, we will guide you through the process.

I know the code is made by volunteers. I have no problem about the time to do things. I was just wondering what are the blocking steps to do such a thing. 

You make all very great work !

I try to contribute when I have a little free time.

romani:
@Calixte , please be welcome with PR.

romani:
fix is merged

romani:
@rnveach , after release is done:
![Screenshot-8 22-from-update-site](https://user-images.githubusercontent.com/812984/68395078-e96ded00-0123-11ea-90a3-9a4006bd96d2.png)
And I see violation from OrderedProperties, as I shown in PR images.

PR #180, opened 2019-11-06
Issue #163: update to Checkstyle 8.22

Calixte:


romani:
@ManfredTremmel , please review

romani:
@rnveach , prove that some new functionality works:
Screenshot from 2019-11-07 05-43-35
Screenshot from 2019-11-07 05-43-36

@Calixte , can you do same prove that new Checks are actually working ? (in next PRs)
we do the same at sonar plugin - https://github.com/checkstyle/sonar-checkstyle/pull/247
So it good practice to recheck and prove reviewers that it works.

PR #198, opened 2020-01-02
remove "Branding" from feature name in about dialog

Bananeweizen:
If an Eclipse feature specifies a branding plugin, then the icon, name
and some other meta data of that plugin are shown as "branding" of the
feature in the Eclipse about dialog. Therefore that specific plugin
should not be called "... branding".

Without this change, the about dialog shows the following: 
image


romani:
@Calixte , please do final review.

PR #200, opened 2020-01-07
fix test classpath

Bananeweizen:
Two of the bundles contain nested tests. However, those have never been
available in Eclipse itself (only in Maven), since they are not on the
classpath of the plugins.

With this change the tests are shown as a separate test source folder
and can be run directly from eclipse. Also potential compiler errors in tests will now be visible immediately (while they were only visible in the maven build output before).

It is not possible to accidentally mix test and main sources, because Eclipse has a main/test classpath
separation (that's the additional attribute "test" with value "true" in the classpath entry).

image

### DIFF
--- a/.settings/org.eclipse.core.runtime.prefs
+++ b/.settings/org.eclipse.core.runtime.prefs
@@ -1,0 +1,2 @@
+eclipse.preferences.version=1
+line.separator=\n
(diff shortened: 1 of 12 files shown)


Bananeweizen:
@rnveach I've edited the classpath by means of using the "Configure Build Path" dialog, which itself writes the file afterwards. The default line terminator in eclipse is the one of the running OS, therefore doing this from a Windows machine (mine) after someone previously did that from a Linux machine leads to such effects.

To avoid this for future changes, I've also set the eclipse project property for line terminators of new files to Linux line endings in the new version of the PR. Therefore everyone using eclipse for editing some file in this repo will now create further Linux line endings.

romani:
@Calixte , please finalize review.

Calixte:
This branch needs a rebase.

PR #209, opened 2020-01-28
Upgrade Tycho to 1.5.1

Bananeweizen:
This may fix the Travis build with Java 12, since Tycho 1.5 supports
Java 12. See https://wiki.eclipse.org/Tycho/Release_Notes/1.5#ECJ_update

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,8 +20,8 @@
     </modules>
 
     <properties>
-        <tycho-version>1.4.0</tycho-version>
-        <tycho-extras-version>1.4.0</tycho-extras-version>
+        <tycho-version>1.5.1</tycho-version>
+        <tycho-extras-version>1.5.1</tycho-extras-version>
         <tycho.scmUrl>scm:git:git://github.com:checkstyle/eclipse-cs.git</tycho.scmUrl>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     </properties>


pom.xml
```diff
@@ -20,8 +20,8 @@
     </modules>
 
     <properties>
-        <tycho-version>1.4.0</tycho-version>
-        <tycho-extras-version>1.4.0</tycho-extras-version>
+        <tycho-version>1.5.1</tycho-version>
+        <tycho-extras-version>1.5.1</tycho-extras-version>
```
rnveach: According to maven it goes to 1.6 . Any reason to stop at 1.5 ?

Bananeweizen: Yes, Tycho 1.6.0 no longer supports Java 12. https://wiki.eclipse.org/Tycho/Release_Notes/1.6#Drop_support_for_obsolete_modular_JVMs_.2810_and_12.29

Calixte:
I approve this. I think we should get this merged so we don't have to deactivate the jdk12 build on the other pull request.

pom.xml
romani: I think it make sense to upgrade Travis to use openjdk13 , 
plan is to verify build on LTS versions (8, 11) and whatever latest public version of java.

@Bananeweizen , please try to update travis config, if it works latest jdk and latest tycho ... good.

romani:
lets merge this to make CI green, but next PR is appreciated to drop openjdk12 support and use openjdk13.

PR #211, opened 2020-01-30
Use Tycho 1.6.0 and JDK 13 in tests

Bananeweizen:
Use the most recent version of Tycho during builds. Since it no longer
supports JDK 12, also upgrade the build matrix to JDK 13 and document
the Java versions of the build matrix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,22 +18,27 @@ addons:
 install:
   - ""
 
+# JDKs used in builds must be supported by Tycho.
+# Check https://wiki.eclipse.org/Tycho/Release_Notes to find supported versions.
 matrix:
   fast_finish: true
   include:
+    # JDK 8 (minimal supported Java version of eclipse-cs)
     - jdk: openjdk8
       env:
         - DESC="install (openjdk8)"
         - CMD="mvn install && git diff"
 
+    # JDK 11 (Long Term Support version)
     - jdk: openjdk11
       env:
         - DESC="install (openjdk11)"
         - CMD="mvn install && git diff"
 
-    - jdk: openjdk12
+    # JDK 13 (most recent Java version)
+    - jdk: openjdk13
       env:
-        - DESC="install (openjdk12)"
+        - DESC="install (openjdk13)"
         - CMD="mvn install && git diff"
 
 script:
--- a/pom.xml
+++ b/pom.xml
@@ -20,8 +20,8 @@
     </modules>
 
     <properties>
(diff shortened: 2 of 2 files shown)


Bananeweizen:
Can someone tell me, why Travis tried to fetch some Maven plugins from some google mirror (?) or Maven central? That's the first time I work with Travis, and obviously one of the builds managed to fetch the artifacts that the other 2 complained about.

romani:
it is the same as https://github.com/checkstyle/checkstyle/issues/7444

Travis need to be updated to not use google mirror.
Let me do this .... in separate PR.

romani:
@Bananeweizen , please rebase, google mirror is disabled in Travis.

romani:
I rebased forked branch ...

.travis.yml
```diff
@@ -18,22 +18,26 @@ addons:
 install:
   - ""
 
+# JDKs used in builds must be supported by Tycho. Check https://wiki.eclipse.org/Tycho/Release_Notes to find supported versions.
```
romani: please wrap long line to follow limit of 100 on line.
now it cause horizontal scrolling - not good for review.

.travis.yml
```diff
@@ -18,22 +18,26 @@ addons:
...
 matrix:
   fast_finish: true
   include:
+    # JDK 8 (oldest supported Java version of eclipse-cs)
```
romani: `JDK 8 (minimal supported Java version of eclipse-cs)`

.travis.yml
```diff
@@ -18,22 +18,26 @@ addons:
...
     - jdk: openjdk8
       env:
         - DESC="install (openjdk8)"
         - CMD="mvn install && git diff"
 
+    # JDK 11 (Long Term Support version)
```
romani: JDK 11 (Long Term Support version , our next minimal version)

.travis.yml
Bananeweizen: Done. Don't easily notice that on 4K :)

.travis.yml
Bananeweizen: Done.

.travis.yml
Bananeweizen: I've not changed that one. I like documentation to reflect the current state, instead of creating confusion by documenting 2 different "minimal versions".

romani: ok, np

romani:
@Calixte , as you are send reviewer, you can simply merge PR is you agree.
No need to wait for me.

PR #229, opened 2020-05-25
Replace boxing constructors

Bananeweizen:
Boxing constructors of primitive types are deprecated in newer versions
of Java and should not be used.

### DIFF
--- a/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/builder/Auditor.java
+++ b/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/builder/Auditor.java
@@ -176,7 +176,7 @@ public class Auditor {
       project.deleteMarkers(CheckstyleMarker.MARKER_ID, false, IResource.DEPTH_ZERO);
 
       Map<String, Object> attrs = new HashMap<>();
-      attrs.put(IMarker.PRIORITY, new Integer(IMarker.PRIORITY_NORMAL));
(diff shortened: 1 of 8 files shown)


net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/builder/Auditor.java
```diff
@@ -176,7 +176,7 @@ private void handleCheckstyleFailure(IProject project, CheckstyleException e)
       project.deleteMarkers(CheckstyleMarker.MARKER_ID, false, IResource.DEPTH_ZERO);
 
       Map<String, Object> attrs = new HashMap<>();
-      attrs.put(IMarker.PRIORITY, new Integer(IMarker.PRIORITY_NORMAL));
+      attrs.put(IMarker.PRIORITY, Integer.valueOf(IMarker.PRIORITY_NORMAL));
```
Calixte, line 179: Why not just `attrs.put(IMarker.PRIORITY, IMarker.PRIORITY_NORMAL);`?
Auto-boxing should take care of this.

Bananeweizen: I must admit I only did a search and replace without really looking at the code, so there might be even more places where that could be optimized further. My main intention was getting rid of all the maven compiler warnings.

Also I'm used to avoid auto-boxing. In my company auto-boxing gives you a compile error, to enforce people to think about boxing/unboxing (and especially about the type and size being used), and I actually like that. :)

PR #235, opened 2020-06-01
avoid flickering during filter update

Bananeweizen:
When updating the list of modules matching the given text filter,
disable redrawing the underlying OS control. Otherwise there is a
flickering effect, since the control is redrawn for every item being
processed by the filter by default.

With redrawing disabled, the control is only drawn once all items have
been filtered, and there is no more flickering.


Calixte:
Interesting. I can't reproduce the flickering.

Bananeweizen:
filter

Calixte:
Do you know why it does not flicker on my macOS or my Ubuntu? Is it only a Windows issue?

Bananeweizen:
I've never been running Eclipse on another OS than Windows myself. But since there are so many platform specific quirks, workarounds etc. in SWT, it seems completely reasonable to me that this might behave differently on other platforms.

Issue #251, opened 2020-08-19
Try to integrate patch-filter to eclipse-cs


HuGanghui:
Try to integrate patch-filter to eclipse-cs.

Reports on how such fitelters works on different commits - https://github.com/checkstyle/patch-filters/issues/297#issuecomment-674979677

romani:
fix is merged.

PR #259, opened 2020-08-29
Issue #251: use released version of patch filter

romani:
Issue #251


romani:
@Calixte , FYI.

I am merging it as it is very minor, no functional changes

PR #260, opened 2020-08-29
Issue #251: auto generate patch file during maven build

romani:
Issue #251


romani:
@Calixte , FYI.

It is restoration of simple build

PR #263, opened 2020-09-12
pack checkstyle fat from maven cache

qxo:
when upgrade version only need change pom.xml


rnveach:
@qxo Do you plan on finishing this. It looks very interesting.

CI failure is:
> Checkstyle Core Library Plugin ..................... FAILURE 
> org.eclipse.tycho:tycho-p2-plugin:1.6.0:p2-metadata-default failed.: IllegalArgumentException

PR #278, opened 2021-02-01
Issue #269: update to Checkstyle 8.36.2

### LINKED ISSUE #269: upgrade to checkstyle 8.36.2
https://checkstyle.org/releasenotes.html#Release_8.36.2

Calixte:
Fixes #269.

### DIFF
--- a/net.sf.eclipsecs.checkstyle/.classpath
+++ b/net.sf.eclipsecs.checkstyle/.classpath
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
-	<classpathentry exported="true" kind="lib" path="checkstyle-8.36.1-all.jar" sourcepath="checkstyle-checkstyle-8.36.zip"/>
+	<classpathentry exported="true" kind="lib" path="checkstyle-8.36.2-all.jar" sourcepath="checkstyle-checkstyle-8.36.2.zip"/>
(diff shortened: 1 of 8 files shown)


romani (approved):
if it works ... lets merge it.
We usually in plugins as to provide some prove that plugin executed and can validate something.
Example: https://github.com/checkstyle/sonar-checkstyle/pull/352 

Travis failure was unrelated to update (it was some connection issue), restarted.

Calixte:
Since the switch to using metadata from the checkstyle jar, this jar is included twice in the plugin.
The plugin is now 27MB instead of 14.5MB. I am trying to solve that issue before releasing more versions.

romani:
>The plugin is now 27MB instead of 14.5MB.

I think in second decade of 21 century it is not a big problem. You can reduy it back in further releases.
I think users are more affected by no sync of IDE and build tool checkstyle version.

famod:
> I think users are more affected by no sync of IDE and build tool checkstyle version.

👍 

We are currently setting up a new project and had to realise that eclipse-cs has fallen behind the cs releases by many versions.
(I know, everyone can contribute but personally I'm already doing so with a couple of other OSS projects.)

IMO, the plugin size problem should be tracked separately.

dkfellows:
Blocking deploying the fix for checkstyle/checkstyle#4638 — which is very noisy in lambda-heavy code — because of a minor file size issue? And with a contributor who has basically disappeared for a month (according to github statistics)? Hmm…

famod:
Just wanted to say thank you @Calixte for all these updates!

Issue #279, opened 2021-02-04
bintray will shut down - update site needs to be moved


Bananeweizen:
The plugin is currently hosted on bintray, which is going to be terminated: https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

-----

Hi Roman,

Thank you for supporting JFrog through your use of Bintray. We are and continue to be a community-focused company, and in the spirit of transparency we’re giving you advance notice that as of May 2021, we will be sunsetting Bintray.

Many of the services that Bintray provides are now served by the JFrog Platform and this sunset allows us to focus our development and innovation efforts on building a comprehensive end-to-end solution for hosting and distribution of software artifacts.

To help minimize any potential disruption in service, we have several solutions for you to employ during this transition period to keep your project running smoothly. Below are details on who is affected, how to mitigate any disruption and additional resources for you to contact if you encounter any snags along the way.

Repositories Hosted on Bintray
Your repositories for module types including Docker, Maven (Java), Npm, NuGet, RPM, Vagrant, and Generic will need to be migrated off of Bintray service on or before May 1st to avoid any disruption in service.
Conan OSS Users
Conan packages for C/C++ have been migrated to ConanCenter.

**Migrate to the Free Subscription of Artifactory**

JFrog has a new free subscription tier that gives you the ability to manage and distribute artifacts with automated security scanning. You can sign up for a free subscription http://mkto-lon040102.com/n/yxNFt8oE0403ZNu00000010 and start migrating your Bintray artifacts. Unlike Bintray, this service also offers Continuous Integration and is suitable for hosting CI artifacts, such as Snapshot builds.​
For additional details, please read the community announcement http://mkto-lon040102.com/n/AN0100FN10E4xo80030v0tZ .

JFrog wants to thank you again for your continued support. We’re committed to our community and strengthening our offerings. Other questions? Please feel free to contact us on Twitter 

Best Regards,
JFrog Team

lkoe:
That's a proper kick in the groin 😬 
Anyone know good alternatives for hosting binaries/p2 repos?

Maven Central is not an option as because p2 repos work with Maven GAV coordinates.
Sourceforge was to unreliable - thats why we migrated to bintray in the first place.
Github pages isn't really meant to host large binaries.
...

Bananeweizen:
Looking at all the plugin URLs that I have in different Oomph setups, the order of occurrences is
* eclipse.org
* github.io
* project's own website
[...]

PR #290, opened 2021-05-04
Issue #279: deploy to github pages instead of bintray

Calixte:
Followed this guide: http://www.lorenzobettini.it/2021/03/publishing-an-eclipse-p2-composite-repository-on-github-pages/
P2 GitHub Pages repository looks like: https://github.com/Calixte/eclipse-cs-update-site
Update site: https://calixte.github.io/eclipse-cs-update-site/

I've tested and it works well.
We probably want to put this repository in the checkstyle organization, and point a better domain to it.


romani (approved):
Better to use org pages, because some other can help you do release.
Let me know if you need access to some repo or ... .

Calixte:
I don't have enough permissions to create a repo at `checkstyle/eclipse-cs-update-site`.
The update site URL would be `checkstyle.org/eclipse-cs-update-site` which is pretty nice.

romani:
@Calixte ,
https://github.com/checkstyle/eclipse-cs-update-site is created, you have Maintainer role.
Let me know if you need anything else.

lkoe:
Nicely done to get it working again.
A small point: you could have just updated the URLs in `docs/update/compositeArtifacts.xml ` and `docs/update/compositeContent.xml` instead of deleting them. This way the public update site url would have stayed as is, without users noticing and forcing them to uninstall/install from the new update site to upgrade.

Calixte:
I'll try to do that for sure. I deleted them because I couldn't find a solution fast enough, and didn't want the old URL to be a source of confusion. 
I also want to bring back the archives that can be manually installed.
And I think it's also missing a reference to the sventu update site.

romani:
We can think on sevntu later on. It has separate update site so users have workaround

PR #303, opened 2021-07-30
Update to Tycho 2.1.0

vogella:
This is the last version which compile the neon target (which is based on Java 1.6, any newer version would require to move the target platform up.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,7 @@
     </modules>
 
     <properties>
-        <tycho-version>1.6.0</tycho-version>
+        <tycho-version>2.1.0</tycho-version>
         <tycho.scmUrl>scm:git:git://github.com:checkstyle/eclipse-cs.git</tycho.scmUrl>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <maven.checkstyle.plugin.version>3.1.1</maven.checkstyle.plugin.version>


Calixte:
Doesn't Tycho 2 require Java 11?
Are you able to run `mvn clean test package` successfully?

Bananeweizen:
It requires Java 11 to run itself, you can still compile code with older Java versions fine. Nevertheless, 1.7.0 is the highest version that I can use to build this repo. Beyond that, the target platform resolution fails, complaining about not being able to resolve the Java 1.6 dependency of the Juno SDK.
That being said, I'm also in favor to increase the target platform to a version that requires Java 8. Compatibility is nice, but still supporting an 8 years old version is really not necessary anymore.

vogella:
+1 for increasing the target platform to a version which requires Java 8.

rnveach:
I am not seeing how this repo can't support anything older than Java 8. Checkstyle requires a Java 8+ runtime for some time. Older JREs won't run with it.

Calixte:
I was not aware you could build eclipse-cs with JDK 7. The [travis configuration](https://github.com/checkstyle/eclipse-cs/blob/master/.travis.yml) would test that we could build it with 8, 11, and 13. 
@vogella do you want to increase the target platform as you update to Tycho 2?

vogella:
@Calixte can do, to which target release should I move the target platform to?

rnveach:
If there is no reason it can't be done, I would split the PRs between Tycho update with the change in target release.

Bananeweizen:
Contained in #353.

Issue #316, opened 2021-09-01
eclipsecs-sevntu-plugin is broken since 8.36.1


rnveach:
Identified at https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/857 ,

It seems eclipsecs-sevntu-plugin's old meta data XML files no longer work, even though eclipse-cs still has code to support it.
https://github.com/checkstyle/eclipse-cs/blob/100aaca6027895c5b1562f4546ee618e04b8a731/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/config/meta/MetadataFactory.java#L568

Installing a version of eclipse-cs on or before 8.35 seems to resolve the issue and bring back the functionality. This seems to lead to https://github.com/checkstyle/eclipse-cs/issues/256 to be the culprit.

PR #314, opened 2021-09-01
Issue #256: trim back old meta data files to rely on checkstyle

rnveach:
Issue #256

This is the first of the attempts to remove the old meta data files. This based on comment at https://github.com/checkstyle/eclipse-cs/issues/256#issuecomment-907832411 .

Currently I cannot remove the XML and property files completely. Doing so either causes exceptions or causes the checks pulled from checkstyle to disappear.

When this is merged, https://github.com/checkstyle/eclipse-cs/issues/113 can be closed as it no longer applies.

Issue #256 can only be closed when the XML/property files are completely gone showing we are completely reliant on Checkstyle.

PR #317, opened 2021-09-01
Issue #316: Allow third party checks to have a custom group name

rnveach:
Issue #316

This seems to resolve the issue for me. I would need an explanation on how to verify the configuration menu works as intended, especially with modules with properties from sevntu.

I believe the issue is we are preventing third parties from creating their own custom group. Sevntu names it's group `SevNTU checks`. We were only allowing names `Other` in.

This may have been intended to prevent third parties from adding their own checks to Checkstyle's main groups?

Screenshot of test usage:
eclipse-cs

Screenshot clearly shows highest eclipse-cs version and sevntu checks with descriptions. I was concerned that nothing was showing on the right hand side of the dialog under "Configured modules for group" and wanted to know how to test this more.

### DIFF
--- a/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/config/meta/MetadataFactory.java
+++ b/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/config/meta/MetadataFactory.java
@@ -723,9 +723,7 @@ public final class MetadataFactory {
       }
 
       // process the modules
-      if (OTHER_GROUP_NAME.equals(groupName)) {
-        processModules(groupEl, group, metadataBundle);
-      }
+      processModules(groupEl, group, metadataBundle);
     }
   }
 


rnveach:
This change seems to have brought back duplicate check names. This may rely on https://github.com/checkstyle/eclipse-cs/pull/314 .
eclipse-cs

rnveach:
The rebase did resolve the double list of checks.

The list of checks are appearing and testing with checkstyle still works, I can produce violations and such.

Switching to Sevntu testing, the list of checks are appearing, not noticing any duplicates. I installed the latest release for sevntu, `1.40.0`. I was able to import a config and have it green check for the sevntu check that is in my config.

However, when switching to violation testing I could not produce a violation from sevntu on any checks (tried 2). I tested with the CLI and could got a violation. I tried switching back and forth with sevntu and checkstyle and only issues with sevntu.

````
$ cat TestClass.java
package test;

/** Main2. */
public class Main2 {
  /** main. */
  public static void main(String[] args) {
    switch (1) {
      case 0:
        switch (2) {
          case 1:
            break;
          default:
            break;
        }
        break;
      default:
        break;
    }
  }
}


$ cat TestConfig.xml
<?xml version="1.0" ?>

<!DOCTYPE module PUBLIC
        "-//Checkstyle//DTD Checkstyle Configuration 1.2//EN"
        "https://checkstyle.org/dtds/configuration_1_2.dtd">

<module name="Checker">
  <module name="TreeWalker">
    <module name="com.github.sevntu.checkstyle.checks.design.NestedSwitchCheck">
    </module>
  </module>
</module>


$ java -jar checkstyle-8.40-sevntu-1.39.0-all.jar -c TestConfig.xml TestClass.java
Starting audit...
[ERROR] TestClass.java:9:9: Nested switch should be avoided, extract nested switch block into separate method. [NestedSwitch]
Audit done.
Checkstyle ends with 1 errors.
````

rnveach:
@Calixte
I don't understand what is wrong with getting a violation with sevntu.
Is there an easy way to debug this issue to find out where the disconnect is?

Though I can't test out violations, the original intention of this PR does succeed in bringing back all sevntu's checks. It just doesn't close the issue if violation isn't working.

rnveach:
@romani @Calixte Please review.

rnveach:
@romani @Calixte ping

Calixte (approved):

romani (approved):

PR #350, opened 2022-05-28
Multiple source cleanups

Bananeweizen:
I was going to fix some deprecations, but doing so I noticed that the code would get littered with more and more duplicated catch-clauses due to newly introduced exception types in the non deprecated methods. Therefore I converted catch to multi-catch and ran some other cleanups. All changes were done fully automated, except for removing some no longer necessary @SupressWarning statements.

### DIFF
--- a/config/checkstyle_checks.xml
+++ b/config/checkstyle_checks.xml
@@ -202,6 +202,15 @@
             <property name="ignoreSetter" value="true"/>
             <property name="setterCanReturnItsClass" value="true"/>
         </module>
+        <module name="IllegalInstantiation">
+            <property name="classes"
+                      value="org.xml.sax.SAXException, org.xml.sax.SAXParseException,
+                         org.apache.commons.beanutils.ConversionException,
+                         org.antlr.v4.runtime.misc.ParseCancellationException,
+                         antlr.RecognitionException, antlr.TokenStreamException,
+                         antlr.TokenStreamRecognitionException, antlr.ANTLRException,
+                         java.lang.StringBuffer"/>
+        </module>
         <module name="IllegalThrows"/>
         <module name="IllegalTokenText">
             <property name="tokens" value="STRING_LITERAL"/>
--- a/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/builder/Auditor.java
+++ b/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/builder/Auditor.java
@@ -499,7 +499,7 @@ public class Auditor {
 
       final String message = error.getMessage();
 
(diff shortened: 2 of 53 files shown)


Bananeweizen:
Can someone explain why the build failed? It looks like it checks all changed files for checkstyle violations. If so, the current EclipseCS sources have around 1400 violations, therefore any somewhat bigger change will fail this check.

Bananeweizen:
This needs to be rebased after #351 has been fixed and the resulting checkstyle issues are also fixed.

rnveach:
> Can someone explain why the build failed? 

https://app.travis-ci.com/github/checkstyle/eclipse-cs/jobs/571713439#L456
These changes resulted in `34 errors` by the rules defined in your configuration at https://github.com/checkstyle/eclipse-cs/blob/5b295acc79fd1bc37e38d9a2f06c01667c57ed29/config/checkstyle_checks.xml#L6 .

Checkstyle checks all files and has no "file changed" recognition. I assume the 1400 errors may be from running with a different configuration then eclipse-cs'.

Travis run seems correct to fail this PR as everything passed at the other PR, https://github.com/checkstyle/eclipse-cs/pull/349 .

romani:
on this branch, on my local, with `mvn install` in root folder of repo, I also have no violations.

romani:
@Bananeweizen ,

Patch filter works on last commit changes.
on local history is linear, so patch using your last commit about `infra: re-enable IllegalInstantiation check`
in Travis, all commits are combined as one merge commit that is applied over master HEAD (it is non configurable feature of Travis) to not execute CI on old code and clearly know result after merge.

So Travis is right :).
you suppose to reproduce this problem is you checkout to each commit seprately and run "mvn install" on each of them

romani:
@Bananeweizen , reminder one more time.
if there be situation with **avalanche of demands** from Checkstyle , **please do not follow it**, report it reviewers and we all can agree to merge with red CI. After merge, it might not be problem, if last commit is not touching problematic lines. So violations will stay in backlog and will not bother people until next update.

romani:
@Bananeweizen , can we finish this PR ?

romani:
I merging this PR as we found problem with Windows for patch filter
https://github.com/checkstyle/patch-filters/issues/113
https://github.com/checkstyle/eclipse-cs/pull/395

without such fix, it will be not easy to fix problems in such big update.
We will keep enforcement in next PRs.

Bananeweizen:
Sorry for not coming back to this earlier. I had tried to debug into the patch filter to understand what's going on, but I failed miserably with that. Next time I'll try to clean up open PRs first.

PR #353, opened 2022-05-30
infra: update maven dependencies and plugins

Bananeweizen:


Bananeweizen:
Be aware the change for JDK 13 to 14 on Travis is a direct consequence of the Tycho 1.6 to 1.7 update. In the future we should build on Java 11 and 17, but that would require Tycho 2.0+, and that in turn should have the Java 11 changes from #349 first. So this change upgrades as far as it can, and further Tycho and JDK upgrades should only come after #349 was implemented.

Bananeweizen:
This now uses Java 11 and 17 for testing, and Tycho 2.7.5 for building. That's almost recent.

Bananeweizen:
Just one additional warning: After this gets merged, you cannot easily build other Tycho projects with versions < 2.7 anymore without manually cleaning some cache in between: https://github.com/eclipse-tycho/tycho/blob/master/RELEASE_NOTES.md#caution-when-switching-between-tycho-versions. However, as long as you build only this one project (and don't switch to old branches), everything will be fine.

rnveach (approved):
If you want assistance just keeping versions up to date, then you may want to look at [dependabot](https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/.github/dependabot.yml). While it isn't perfect, it atleast makes things a bit easier.

Calixte (approved):

Issue #360, opened 2022-08-17
Broken link in README.md for obtaining older versions


andrewsgoetz:
README.me says "Older versions can be installed using the specific update site for that version, for instance https://checkstyle.org/eclipse-cs-update-site/8.41.0/".

However, the sample link is broken, and it is also broken for other different version numbers that I've tried (9.0.0, 10.0.0).

romani:
What do mean by broken? It should not be open in browser, it should work only in Eclipse

andrewsgoetz:
Ah, you're right, I should have read that more closely. I did eventually work out how to install the version of Checkstyle I wanted (9.0.0) in my version of Eclipse (2020-03), but it took me a lot of trial and error to figure out how. I didn't end up using the link https://checkstyle.org/eclipse-cs-update-site/9.0.0/ directly.

I had to go to Help->Install New Software, click on "Add...", and add https://checkstyle.org/eclipse-cs-update-site/. Then I had to uncheck "Show only the latest versions of available software" so that I could see older versions.

I wonder if the readme could be updated to be clearer.

romani:
Please propose new wording

vtintillier:
@andrewsgoetz the links that work for me are for example
- http://checkstyle.org/eclipse-cs-update-site/updates/8.x/8.44.x/
- https://checkstyle.org/eclipse-cs-update-site/updates/9.x/9.3.x/

Bananeweizen:
Thanks for the PR.

PR #361, opened 2022-08-18
Update installation instructions.

andrewsgoetz:
Pull request coming from discussion here: https://github.com/checkstyle/eclipse-cs/issues/360.


Bananeweizen:
Looks good to me. I also think that most users prefer the marketplace wizard instead of using drag and drop, so it's fine to replace that.

Issue #322, opened 2021-09-21
Magic Numbers rule UI cannot be opened (NPE)


MathiasAugstein:
I cannot add Coding_Problems/MagicNumber rule to our configuration. Using the "Add... ->" Button has no effect, I get the error:
```
java.lang.NullPointerException
	at net.sf.eclipsecs.ui.config.widgets.ConfigPropertyWidgetFactory.createWidget(ConfigPropertyWidgetFactory.java:57)
	at net.sf.eclipsecs.ui.config.RuleConfigurationEditDialog.createConfigPropertyEntries(RuleConfigurationEditDialog.java:518)
	at net.sf.eclipsecs.ui.config.RuleConfigurationEditDialog.createGeneralSection(RuleConfigurationEditDialog.java:173)
	at net.sf.eclipsecs.ui.config.RuleConfigurationEditDialog.createDialogArea(RuleConfigurationEditDialog.java:127)
	at org.eclipse.jface.dialogs.TitleAreaDialog.createContents(TitleAreaDialog.java:162)
	at org.eclipse.jface.window.Window.create(Window.java:431)
	at org.eclipse.jface.dialogs.Dialog.create(Dialog.java:1094)
	at net.sf.eclipsecs.ui.config.RuleConfigurationEditDialog.create(RuleConfigurationEditDialog.java:529)
	at org.eclipse.jface.window.Window.open(Window.java:788)
	at net.sf.eclipsecs.ui.config.CheckConfigurationConfigureDialog$PageController.newModule(CheckConfigurationConfigureDialog.java:708)
	at net.sf.eclipsecs.ui.config.CheckConfigurationConfigureDialog$PageController.widgetSelected(CheckConfigurationConfigureDialog.java:514)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:252)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:89)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4237)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1060)
	at org.eclipse.swt.widgets.Display.runDeferredEvents(Display.java:4054)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3626)
	at org.eclipse.jface.window.Window.runEventLoop(Window.java:823)
	at org.eclipse.jface.window.Window.open(Window.java:799)
[...]

Issue #351, opened 2022-05-29
Inconsistent checkstyle rules in the project itself


Bananeweizen:
The Eclipse plugins are currently configured to use the Google checks, see for instance https://github.com/checkstyle/eclipse-cs/blob/master/net.sf.eclipsecs.core/.checkstyle. However, during a PR build, the (otherwise not referenced) configuration from https://github.com/checkstyle/eclipse-cs/blob/master/config/checkstyle_checks.xml is used. That leads to the local IDE being free of errors, but pull requests failing, as I experienced myself.

I could create a project relative configuration to use that shared configuration file, but there are multiple issues:
* It has 7 properties and those would need to be defined to make EclipseCS be able to use it. Those properties must be set during the build somehow, does anyone know where and how?
* Using this project relative configuration requires that every EclipseCS developer imports the root level Maven project into the Eclipse workspace. Unfortunately it is not possible to specify a relative configuration location without a containing project.
* That shared config file looks like a copy of the file from the main Checkstyle project. It should probably be updated, or be referenced from there directly. Similar to the previous item, this would require every EclipseCS developer to have the root project of the Checkstyle repo be imported in the Eclipse workspace.

Once this configuration has been updated, a lot of errors need to be expected. Those should be fixed before doing any further Java work. Otherwise every PR has the chance of failing with Checkstyle errors as happened here: https://github.com/checkstyle/eclipse-cs/pull/350

rnveach:
IMO, I would start by updating https://github.com/checkstyle/eclipse-cs/blob/master/config/checkstyle_checks.xml to be the configuration you wish to be use for all of eclipse-cs. It is already controlled by CI, so the only thing left would be to configure Eclipse to use it.

I would be willing to assist with violations once this configuration is defined.

For the main repository, we have a few sub-repos that also use the main's configuration file so they all use the same configuration. I don't know eclipse-cs well enough if it could make use of it, but we define it in maven via the following items:

Specify the external configuration location in the POM:
https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/pom.xml#L22
Values to the repo-specific properties used inside the config:
https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/checkstyle.properties

> It has 7 properties and those would need to be defined to make EclipseCS be able to use it. Those properties must be set during the build somehow, does anyone know where and how?

[...]

PR #376, opened 2022-11-06
Issue #322: NPE on rule configuration

Bananeweizen:
Issue #322

Broken since introduction of tokenTypesSet in
https://github.com/checkstyle/checkstyle/commit/d038f187190ee5c9d2d55b6aae985b3ec9327614 of the main checkstyle project.

### DIFF
--- a/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/config/meta/MetadataFactory.java
+++ b/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/config/meta/MetadataFactory.java
@@ -73,6 +73,11 @@ import org.yaml.snakeyaml.Yaml;
  */
 public final class MetadataFactory {
 
+  /**
+   * metadata type or validation type for regular expressions
+   */
+  private static final String TYPE_ID_PATTERN = "java.util.regex.Pattern";
+
   /** Map containing the public - internal DTD mapping. */
   private static final Map<String, String> PUBLIC2INTERNAL_DTD_MAP = new HashMap<>();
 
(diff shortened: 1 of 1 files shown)


rnveach:
Is there something preventing adding tests for this code to ensure it is not broken in the future?

Bananeweizen:
> Is there something preventing adding tests for this code to ensure it is not broken in the future?

Not really sure where to start adding tests there, that's an area of code I haven't been exploring before. I was rather thinking of adding more assertions for nonnull members for all the config classes, which would have detected this problem at the moment of using the new checkstyle version.

The failing tests are caused by #351, which is why I have finally taken time to fix that. All other open failing PRs would need to be rebased after that is merged.

rnveach:
https://app.travis-ci.com/github/checkstyle/eclipse-cs/jobs/587945887
Failure is the same as https://github.com/checkstyle/eclipse-cs/pull/377#issuecomment-1309710185 and should be fixed as soon as we release patch-filters and upgrade to it.

romani:
Failure was
```
Caused by: java.lang.NoSuchMethodError: 'com.puppycrawl.tools.checkstyle.api.LocalizedMessage com.puppycrawl.tools.checkstyle.TreeWalkerAuditEvent.getLocalizedMessage()'
	at com.puppycrawl.tools.checkstyle.filters.JavaPatchFilterElement.isLineMatching(JavaPatchFilterElement.java:217)
```

Should be resolved after merge of https://github.com/checkstyle/eclipse-cs/pull/383 with upgrade of patch-filter to 1.2.0

romani:
Travis failure
```
[INFO] --- maven-checkstyle-plugin:3.2.0:check (sevntu-checkstyle-check) @ net.sf.eclipsecs.core ---
[INFO] There is 1 error reported by Checkstyle 10.0 with config/checkstyle_sevntu_checks.xml ruleset.
[ERROR] src/net/sf/eclipsecs/core/config/meta/
MetadataFactory.java:[314,35] (extension) MultipleStringLiteralsExtended: 
The String "java.util.regex.Pattern" appears 2 times in the file.
```

Did you try to fix it? Does it avalanche of other complains from checkstyle?

romani (approved):
we need to agree on some way of sharing how it was tested (as we have no auto tests).
@Bananeweizen , please suggest a way.

for CS version bump commits, we agreed to share images that new Check/propertties is rendering  fine in UI and can produce violations.

Bananeweizen:
I've reproduced the NPE by just adding the MagicNumber check to the list of active checks in the configuration UI. With this change, the NPE no longer happens. In fact, all checks can be added via the configuration UI now, without causing any exception. I'd rather like to write such a short sentence than posting some images, since images don't tell anything how to reproduce the testing.

romani:
sounds like most easy way to reproduce a problem..... not sure why @Calixte did not run into it while shallow-testing of new CS version updates.

Bananeweizen:
Well, everything worked okay if you would only try configuring something from the top of the list of checkers. Therefore I can understand he didn't notice. In fact, I also didn't run into the issue ever before, and I also did similar quick tests several times. (and at work with a real life config, I use the XML editor directly)

PR #378, opened 2022-11-06
Use the shared configuration for both CI and IDE

Bananeweizen:
Until now the local Eclipse IDE containing the EclipseCS sources would not actively check its own source code, therefore the CI builds on github raised errors for changes that were fine locally. Configure the eclipse plugins to also use that shared configuration instead of some Google Checks (for this also the gitignore had to be cleaned up).

Afterwards remove all checks from the shared configuration which are currently violated. Otherwise there are 7000 errors in the code. When this has been merged, the removed changes can be re-added one by one and the violations can be fixed (that's why there is a separate commit for that file). Otherwise every normal bugfix PR would be blocked further for months due to violating the rules in the PR build without seeing any problem locally.


Calixte (approved):

config/checkstyle_checks.xml
```diff
@@ -12,6 +12,15 @@
         <property name="basedir" value="${basedir}"/>
     -->
 
+    <!--
+    This is a configuration copied from the main Checkstyle project.
+    Unfortunately the Eclipse-CS code hasn't followed its own principles for a long time.
+    Therefore this configuration has been stripped down to get rid of any warning first.
+    The plan is to re-add check by check, together with the necessary fixes in the Java code.
+    However, that also requires to first align the Eclipse formatter settings and the
+    Checkstyle configuration and then to reformat all the sources once.
```
rnveach, line 21: @Bananeweizen @Calixte Do we have an issue to track this on? It doesn't seem positive that we disabled so many possible useful checks.

If the plan is to eventually use the same configuration as checkstyle, then it is easier to just use it's URL to pull in the file.
Example: https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/pom.xml#L22

Bananeweizen: @rnveach Even reducing the scope to modified files is not sufficient (looking at my experience in other projects). If I change 2 lines in a file for a bug fix, but then get another 10 checkstyle issues on the remaining 200 lines, the net result is
* I have to think about 10 (often different!) further issues and their potential fixes
* Really cleaning up those 10 issues can easily lead to further files being modified, thereby more fixes to be requested from the CI system
* My motivation to actually contribute a bug fix in the first place gets negative.

On the other hand, just enabling one more rule in a totally separate PR and fixing its 100 occurrences is way more simple to do (always the same fix), and has higher impact than the above "mixed-in" fixes, leading to cleanining up the code base faster. I have done this kind of "start with empty rules, add one by one back with fixes" several times already in 6 digit LOC code bases, and it was way easier than "take a full rule set, fix everything on change".

Fixing the formatter first has this reasoning:
* Right now the code doesn't have a unique formatting applied, but several "personal" styles, leading to "clusters" of violations.
* Fixing that is most simple by not editing line by line, but by applying the eclipse formatter to all existing files.
* To avoid multiple formatting styles in the future to come back, the formatter settings need to be checked in with the project instead of relying on the IDE workbench global settings. Every developer might have modified those according to personal taste. Many Eclipse developers are not aware of that difference (formatter settings in project properties versus formatter settings in workbench) and never check in those settings unfortunately.

rnveach: >  If I change 2 lines in a file for a bug fix, but then get another 10 checkstyle issues on the remaining 200 lines, the net result is

My understand of patch filter is this is not how it behaves. If you change 2 lines, it only allows those 2 lines through for Checkstyle to report violations on.

> enabling one more rule in a totally separate PR and fixing its 100 occurrences is way more simple to do 

I agree, but just saying with patch filter enabled, you wouldn't even see those 100 occurrences. I am assuming if a PR only has a configuration change, it will never display any violations even if new checks are added or properties are changed.

Sevntu has it in place still (which is why it failed without a patch filter upgrade) and there are 500+ violations being hidden from your view. You should see this in your current PRs.

====

Regardless, we still need a new issue to track the re-enablement. When you have time, please create this issue since you understand it all best.

rnveach: > Fixing that is most simple by not editing line by line, but by applying the eclipse formatter to all existing files.

I can see this for some things yes, but not for all. I am not aware of an Eclipse setting that can add headers. OrderedProperties probably has to be done manually.

I can see other things being too complex in one go and I would avoid like the plague, like adding Javadocs, which again Eclipse can't add for you automatically.

If my previous post alleviates your concerns about patch filter's functionality, I suggest we enable the configuration back in after we are on the latest checkstyle (10.4), synch the main repo's configuration so we are all the same, and add patch filter to work on the violations slowly. If patch filter doesn't do as expected, then we create an issue for it and disable things as we go.

romani:
@Bananeweizen, @Calixte , please pass through me and rnveach next time such updates, following a checkstyle is our religion. We must eat food we produce.

As I mentioned in other PR, if fixing checkstyle issues cause avalanche , it should be raised and we can approve merge with checkstyle violations. 

But disablement of Checks is not good. Please restore modules. 

Yes, project is migration to better and unified style, but we should not disable Checks, it will cause problems to grow.

PR #379, opened 2022-11-08
infra: Use JDK 11 and 17

Bananeweizen:
I haven't found any documentation on whether travis has jdk 17 installed. But if so, then we should use it to test on the LTS versions that are actually supported by our plugin.


Bananeweizen:
have to get the other maven updates in first...

PR #380, opened 2022-11-09
infra: disable sevntu checks on eclipse-cs

Bananeweizen:
All my recent PRs fail on Travis when running the sevntu checks. Locally the same checks pass. Until someone can figure out why this breaks on Travis, it should remain disabled. Otherwise no bug fixes can be merged.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -173,7 +173,13 @@
                     </execution>
                     <execution>
                         <id>sevntu-checkstyle-check</id>
-                        <phase>verify</phase>
+                        <!--
+                        FIXME: The sevntu checks fail basically every build on Travis.
+                        However, locally the same checks pass. Until someone figures out
+                        why that fails on Travis, the checks should remain disabled,
+                        so that we can merge some PRs.
+                        -->
+                        <phase>none</phase>
                         <goals>
                             <goal>check</goal>
                         </goals>


rnveach:
@Bananeweizen Can you point out failures and where they started?

https://app.travis-ci.com/github/checkstyle/eclipse-cs/builds
All travis builds are green.

rnveach (changes requested):
I looked around and only saw PR failures. I confirmed in my own PR that master is passing still, so I can only assume right now that failures you are experiencing are directly related to your own changes.

I would prefer we try to do some investigation as to why the failure is happening instead of just going straight to disabling so we can say what we are looking for to re-enable. If we do end of disabling, we need to create an issue for its resolution.

I will assist in what investigation I can. Please point me to a single place where to start. I made changes at https://github.com/checkstyle/eclipse-cs/pull/381#issuecomment-1309660921 in travis which I hope will help in printing more information why maven is failing for you once it is merged.

rnveach:
Failure in other PRs seem to be related to https://github.com/checkstyle/patch-filters ( https://github.com/checkstyle/eclipse-cs/blob/master/pom.xml#L152-L156 ) and it using an incompatible version with the checkstyle version defined in the maven plugin run. Sevntu is not the issue and the ID of the run is just named sevntu though it includes more.

Once new version of patch-filters is released, we can upgrade it here and hopefully this will resolve the PR issues.

Bananeweizen:
@rnveach Thanks for taking care of this. Another alternative might have been reverting the part of #353, which upgraded the version numbers of checkstyle. However, I can't really judge that, you surely know better. I'll probably just sit back and wait now, until I can rebase the PRs.

rnveach:
You're right, I didn't look into it but that is when it became an issue waiting to happen. I was unaware of patch-filters in this repo, so we need to update the comment above to mention it must be a 3 way match between checkstyle, sevntu, and patch-filters. We should also add patch-filters version to the list so they are all close by the comment. I see a few other changes that should be made.

We can do all that when patch-filters is updated.

romani:
what version of checkstyle we need at patch-filter 
https://github.com/checkstyle/eclipse-cs/blob/0bdcad7bd697049056c83643606be1085588214f/pom.xml#L29

we we need them all on checkstyle 10.0 ?

romani:
I released 1.2.0 version of patch-filter, should be here soon - https://repo1.maven.org/maven2/com/puppycrawl/tools/patch-filters/
please try to use it to fix problem. Pin me if still problem.

rnveach:
@Bananeweizen I believe I have an understanding now why it is only failing in your PRs and not mine or master.

patch-filters is a specialized filter that looks for changes in the git history to know if checkstyle should run and run only on the files that changed. If there are no changes to files Checkstyle validates, it prevents a complete Checkstyle run. If only 1 file changes, then Checkstyle runs only on that 1 file and not the rest, and so on.

All the PRs that were made that did not fail, changed files that Checkstyle does not verify against, so patch-filter prevented a Checkstyle run in those. Your PRs did make a change in files Checkstyle examines, so patch-filters saw that and was attempting to hand those files over to Checkstyle to examine when the exception occurred.

romani (changes requested):
I think this PR should be closed

Bananeweizen:
Thanks for explaining. I still have to figure out how PRs behave after he patch filter becomes active, I have never used that one before in my project.

PR #382, opened 2022-11-10
infra: disable maven download messages

Bananeweizen:
Use the long versions of the options since others reading the file might not know the short versions. If the build fails on the no-progress option, then it needs to be removed. That one is only available in recent maven versions.


rnveach (approved):

romani (approved):

Calixte (approved):

rnveach:
@Calixte Is there something preventing you from merging when all 3 active players agree to PR?

PR #387, opened 2022-11-12
infra: re-enable colored output on Travis

Bananeweizen:
This reverts a part of my previous PR. In early Maven versions the batch-mode option was the only way to suppress download messages. However, since Maven 3.5 it also suppresses the ANSI escape characters that cause colored output in the shell. That's why the current Travis console is no longer colored after the previous commit.

Long story short: On recent maven versions, only the -ntp option should be used, but not the batch-mode option. (I only learned this yesterday evening, otherwise I would have avoided adding it in the first place)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,14 +30,14 @@ jobs:
     - jdk: openjdk11
       env:
         - DESC="install (openjdk11)"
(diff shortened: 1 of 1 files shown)


romani (approved):
ok to merge

PR #389, opened 2022-11-12
ci: replace tycho-source-feature-plugin

Bananeweizen:
The previously separate tycho-source-feature-plugin functionality is included in tycho-source-plugin meanwhile:
https://github.com/eclipse-tycho/tycho/blob/master/RELEASE_NOTES.md#tycho-source-feature-generation-moved-from-tycho-extras-to-tycho-core. Replacing it removes one warning from the Maven build.

Also move the configuration into the parent POM, since it is independent of the specific feature being built (i.e. could be reused if another feature would be built).

And remove some commented out configuration of the p2-metadata plugin. We wouldn't want to use it with those configuration values anyway.

Verified the source feature in the update site is identical to what it was before. BTW: I'm also committer on the Tycho project. :)

### DIFF
--- a/net.sf.eclipsecs-feature/pom.xml
+++ b/net.sf.eclipsecs-feature/pom.xml
@@ -12,25 +12,8 @@
     <build>
         <plugins>
             <plugin>
-                <groupId>org.eclipse.tycho.extras</groupId>
-                <artifactId>tycho-source-feature-plugin</artifactId>
-                <configuration>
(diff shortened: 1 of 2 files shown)


romani (approved):

Bananeweizen:
For anyone interested, this warning from previous builds is gone in the current PR, as expected: 
grafik

Issue #393, opened 2022-11-13
Sort modules alphabetically


Bananeweizen:
Whenever a list of elements in a user interface doesn't have a natural order, then it should be sorted alphabetically for easier reading and lookup:
![grafik](https://user-images.githubusercontent.com/406876/201511311-d57c7506-1db7-4c12-adf2-2f62a619f5e6.png)

romani:
Fix is merged

PR #394, opened 2022-11-13
Issue #393: Sort available modules alphabetically

Bananeweizen:
That makes it easier for an end user to locate a wanted module and to find unknown modules. Also change the label to "Available modules".

grafik


net.sf.eclipsecs.ui/src/net/sf/eclipsecs/ui/messages.properties
```diff
@@ -20,7 +20,7 @@ CheckConfigurationConfigureDialog_lblConfiguredModules = Configured modules for
 
 CheckConfigurationConfigureDialog_lblDescription = Description:
 
-CheckConfigurationConfigureDialog_lblKnownModules = Known modules
+CheckConfigurationConfigureDialog_lblKnownModules = Available modules
```
romani, line 23: is it problem to change key ?
to be `CheckConfigurationConfigureDialog_lblAvailableModules`

romani (changes requested):
items

net.sf.eclipsecs.ui/src/net/sf/eclipsecs/ui/messages.properties
Bananeweizen: No problem. Be aware there are no more blanks in the properties file, because that was changed automatically on refactoring the name of the Java member. Stuff like that is why I argued everything needs to be reformatted once.

romani (approved):

net.sf.eclipsecs.ui/src/net/sf/eclipsecs/ui/messages.properties
romani: please send PR with reformatting all.

config/sevntu_suppressions.xml
```diff
@@ -5,4 +5,8 @@
         "https://checkstyle.org/dtds/suppressions_1_1.dtd">
 
 <suppressions>
+	<!-- Messages.java are generated by the Eclipse localization wizard, therefore field order is generally not managed by human editors -->
+    <suppress checks="CustomDeclarationOrder"
```
romani, line 9: Indentation to fix

romani (changes requested):
Items

config/sevntu_suppressions.xml
Bananeweizen: I guess you are talking about the tab that slipped in at the comment? If so, we should have a checkstyle config for the configuration files. I have that at work to avoid things like duplicate ids, repeated checks etc.

romani: >we should have a checkstyle config for the configuration files

We usually do. But you disable bunch of checks.
So in this repo we prone to such minors.

romani (approved):

config/sevntu_suppressions.xml
rnveach: It is also more than 100 characters.

rnveach:
I assume this would also sort sevntu checks if they were installed? (and working, https://github.com/checkstyle/eclipse-cs/issues/316 )

romani (approved):

Bananeweizen:
> I assume this would also sort sevntu checks if they were installed? (and working, #316 )

Since that is a property of the viewer, it would affect all items that are displayed there. And to my knowledge that means all recognized internal and external checks.

PR #398, opened 2022-11-13
supplemental: fix all Eclipse IDE warnings

Bananeweizen:
This change is aimed at having zero warnings in the Eclipse IDE when working on EclipseCS. Without this change there are around 120 warnings from JDT, PDE and M2E.

* Have all available preferences in jdt.core.prefs and pde.prefs to ensure JDT and PDE compiler settings are version controlled.
* Set most compiler warnings to error to avoid re-introducing any of them. Some checks must be left at warning level, because there are occurrences where we use SuppressWarnings("foo"), and that only works if the level is not error.
* Ignore all deprecation warnings, they are not fixable when we want to remain compatible with old Eclipse versions.
* Fix all remaining warnings.
* Add lifecycle mapping for Eclipse m2e to avoid Maven lifecycle mapping warnings. This has no effect for other Maven users.

For reviewers: The only real code change is in MarkerPropertyPage.java (and has been verified by testing the markers page afterwards). Every other change is one of
* adding some suppression
* rename refactoring (always on non overridden methods)
* removal of duplicate super interface

Should look like this for other EclipseCS developers afterwards:
grafik
If not, I'm happy to investigate further settings fixes.

BTW: This does not yet include any formatter, cleanup or similar preferences, only the compiler warnings preferences are part of this change.


.gitignore
```diff
@@ -1,10 +1,7 @@
 bin/
 target/
-metadata-gen/
```
Bananeweizen, line 3: The metadata-gen doesn't exist anymore since 8.x, when that was refactored to load original Checkstyle metadata.

romani (approved):

net.sf.eclipsecs.branding/.project
```diff
@@ -21,12 +21,12 @@
 			</arguments>
 		</buildCommand>
 		<buildCommand>
-			<name>org.eclipse.m2e.core.maven2Builder</name>
+			<name>net.sf.eclipsecs.core.CheckstyleBuilder</name>
 			<arguments>
 			</arguments>
 		</buildCommand>
 		<buildCommand>
-			<name>net.sf.eclipsecs.core.CheckstyleBuilder</name>
```
Bananeweizen, line 29: If you are wondering: M2E wants to be the last builder generally.

PR #399, opened 2022-11-14
ci: upgrade Tycho to 3.0

Bananeweizen:
That is the most recent version as of today. One of the plugins has been moved from tycho.extras to tycho core, therefore the groupid is different.

Also fix the warnings `[WARNING] artifact
foo:bar:p2metadata:10.3.3-SNAPSHOT already attached, replace previous instance` by disabling the default metadata generation. Unfortunately this extra step is necessary when creating source features, since the attaching of metadata needs to be delayed then.

There are new warnings at the very end of the build like `No digest algorithm is available to verify download`. That's an open issue in Tycho but it doesn't break the build, see
https://github.com/eclipse-tycho/tycho/issues/1482

Finally the target platform also needs to be updated. Tycho 3.0 resolves the JUnit classpath container of the .classpath files against the bundle names introduced in Eclipse 2022-06. Most thirdparty libraries have been changed from repackaging in Eclipse Orbit to "consume as is" from Maven. This does not affect the EclipseCS plugin created by the build, since the new JUnit bundles are only needed at IDE or build time.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -25,11 +25,13 @@ jobs:
         path: ~/.m2/repository
         key: maven-cache-${{ hashFiles('**/pom.xml') }}
 
-    - name: Set up JDK ${{ matrix.jdk }}
+    - name: Set up JDKs
       uses: actions/setup-java@387ac29b308b003ca37ba93a6cab5eb57c8f5f93 # v4.0.0
       with:
         distribution: temurin
-        java-version: ${{ matrix.jdk }}
+        java-version: |
+          11
+          17
 
     - name: Build with Maven
       run: ./mvnw --errors --no-transfer-progress install
@@ -40,6 +42,6 @@ jobs:
     - name: Publish Test Results
       uses: EnricoMi/publish-unit-test-result-action@4e121453199042300b654ee3b4b52720bc645dc6 # v2.13.0
       # we only want to attach test results one time, not for every matrix combination
-      if: startsWith(matrix.platform, 'ubuntu') && (matrix.jdk == 17)
+      if: startsWith(matrix.platform, 'ubuntu') && (matrix.jdk == 11)
       with:
         files: '**/target/surefire-reports/*.xml'
--- a/net.sf.eclipsecs.target/Maven update check.launch
+++ b/net.sf.eclipsecs.target/Maven update check.launch
@@ -18,7 +18,8 @@
         <listEntry value="org.eclipse.debug.ui.launchGroup.run"/>
     </listAttribute>
     <booleanAttribute key="org.eclipse.jdt.launching.ATTR_ATTR_USE_ARGFILE" value="false"/>
+    <booleanAttribute key="org.eclipse.jdt.launching.ATTR_SHOW_CODEDETAILS_IN_EXCEPTION_MESSAGES" value="true"/>
     <booleanAttribute key="org.eclipse.jdt.launching.ATTR_USE_CLASSPATH_ONLY_JAR" value="false"/>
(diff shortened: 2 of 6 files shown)


Bananeweizen:
@rnveach Maybe I could need some help with the Travis configuration for this one later. Tycho 3.0 needs Java 17, even though it can compile and test against Java 8, 11 and 17. I know how to configure this correctly locally, by using a JAVA_HOME pointing to 17 for execution of Tycho and having an /.m2/toolchains.xml pointing to the Java 11 JDK needed for compilation/test. Can the Travis configuration be tweaked similarly? Right now a single JDK is used both for the Maven execution and for the compilation inside, right?

rnveach:
> Right now a single JDK is used both for the Maven execution and for the compilation inside, right?

This is correct as seen at the following locations:
https://github.com/checkstyle/eclipse-cs/blob/master/.travis.yml#L30
https://github.com/checkstyle/eclipse-cs/blob/master/.travis.yml#L36

> Tycho 3.0 needs Java 17
> using a JAVA_HOME pointing to 17 for execution of Tycho and having an /.m2/toolchains.xml pointing to the Java 11 JDK needed for compilation/test.

I found https://github.com/travis-ci/travis-cookbooks/issues/35#issuecomment-6015843 when googling.

I am not sure how possible this will be. I would also be very hesitant on setting up a complex CI run. It is always a pain getting things just right and the littlest things break everything.

Setting the java version in the POM should ensure we compile backwards for Java 11,
https://github.com/checkstyle/eclipse-cs/blob/master/pom.xml#L307-L308
https://github.com/checkstyle/checkstyle/blob/master/pom.xml#L224
https://github.com/checkstyle/checkstyle/blob/master/pom.xml#L1323

but this means no one can compile this project without Java 17 as tycho run is required every maven run. You also have to remember everyone (old and new) will need to set up their local in a similar way otherwise they can't run development.

I am against such a complex change and we should stick to the highest version of tycho that supports Java 11 for now.

@romani

Bananeweizen:
I'm fine with staying on 11, if that leads to complicated CI setups. Honestly, I thought every Java developer has 17 on the local machine meanwhile, so at least that should not be an issue. :) There is no immediate benefit from using Tycho 3.0, except of "update dependencies as often as possible" to find new problems early. I'll port the warnings fix into a separate change then.

romani:
Better to stay on jdk11.

If we want be prepared for jdk17, we can make extra CI execution

Bananeweizen:
_Reviewers, please be aware that this PR has changed a lot since it started. It's now about Tycho 4 instead of 3 and it deals fine with the Java 11 and 17 mix. This is the current commit message:_

That is the most recent version of Tycho. It requires Java 17 for
running, but can compile Java 11 code.

One of the plugins has been moved from tycho.extras to tycho core,
therefore the groupid is different. Some new tycho plugins have been
added. Even though we don't use them, they need to be configured due to
our strict enforcer checks.

There are new warnings at the very end of the build like `No digest
algorithm is available to verify download`. That's an open issue in
Tycho but it doesn't break the build, see
https://github.com/eclipse-tycho/tycho/issues/1482

Finally the target platform also needs to be updated. Tycho 3.0 resolves
the JUnit classpath container of the .classpath files against the bundle
names introduced in Eclipse 2022-06. Most thirdparty libraries have been
changed from repackaging in Eclipse Orbit to "consume as is" from Maven.
This does not affect the EclipseCS plugin created by the build, since
the new JUnit bundles are only needed at IDE or build time.

The jgit exception at the end of a Maven build can be ignored. It's
fixed already upstream, but will probably only be available for use in
Tycho with the Eclipse 2024-03 release.

Bananeweizen:
@rnveach Do you want to give this a second look? The matrix configuration of the github workflow is not an issue anymore, since I install both 11 and 17 in every run. Also requiring Java 17 on the developer machine is a non-issue now, the Eclipse IDE also requires it meanwhile.

Calixte (approved):

PR #400, opened 2022-11-14
ci: fix warnings in maven console

Bananeweizen:
Fix the warnings `[WARNING] artifact
foo:bar:p2metadata:10.3.3-SNAPSHOT already attached, replace previous instance` by disabling the default metadata generation. Unfortunately this extra step is necessary when creating source features, since the attaching of metadata needs to be delayed then.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -233,6 +233,15 @@
                     <version>${tycho-version}</version>
                     <executions>
                         <execution>
+                            <!-- this execution is bound via default lifecycle bindings of the tycho-p2-plugin -->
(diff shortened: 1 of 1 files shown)


Bananeweizen:
Just for reference, I've diagnosed the same problem (and fix) for another project here: https://github.com/eclipse-tycho/tycho/issues/1100

Bananeweizen:
I've verified the log of Travis. The mentioned warning is gone in both builds.

rnveach (approved):

PR #402, opened 2022-11-15
dependency: upgrade snakeyaml to 1.33

Calixte:
I haven't tested this code path yet.
I think it's only used when loading third party checks metadata, so I'd need to install eclipsecs-sevntu-plugin.


rnveach:
> eclipsecs-sevntu-plugin

https://github.com/checkstyle/eclipse-cs/issues/316

Bananeweizen:
Sorry for updating your branch twice. I used the web editor the first time for resolving a conflict (from the 10.4 change), and didn't like the merge commit in between.
To me the change is fine to merge. We probably need to do a more extensive testing of sevntu after that has been upgraded to 10.4 anyway.

PR #404, opened 2022-11-15
infra: fix Maven warning in Eclipse

Bananeweizen:
Eclipse still warns about not being able to map the maven clean goal to the Eclipse build lifecycle. Since that maven goal is not necessary for a successful build, we can ignore it inside Eclipse, similar to the other goals of the resources plugin that have already been ignored.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -360,6 +360,19 @@
                                         <ignore></ignore>
                                     </action>
                                 </pluginExecution>
+                                <pluginExecution>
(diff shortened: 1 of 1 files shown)


Bananeweizen:
Sorry, I confused my browser windows. I totally didn't intend to merge this without a build and review. I'll closely monitor the CI build now, but that change is Eclipse specific and should not lead to any problem.

PR #406, opened 2022-11-15
infra: convert nested libs to wrapped OSGi bundles

Bananeweizen:
EclipseCS uses several external libraries, which are copied manually into the libs folder. Since everything else comes from the Eclipse target platform, those 3 libraries are easier to maintain by fetching them directly from maven and letting Eclipse PDE automatically wrap them into OSGi bundles.

With that change in place, updating the external libraries is as easy as changing the version number in the .target file.

For end users, this change is in fact visible. Previously the nested libs were contained in the checkstyle jar, now they are visible  as plugin in the IDE. This is how the newly wrapped javassist library looks like (that entry did not exist before):
grafik


Bananeweizen:
For reviewers: The interesting parts from the build are the lines around https://app.travis-ci.com/github/checkstyle/eclipse-cs/jobs/588568308#L346, where Tycho mentions how it wraps one of the libs from maven as Eclipse plugin.

rnveach:
Please show that you did testing and Checkstyle still functions correctly.

> are easier to maintain by fetching them directly from maven 
> it wraps one of the libs from maven as Eclipse plugin

This sounds great that the JARs are being removed. Why isn't this also possible with the Checkstyle JAR?

> updating the external libraries is as easy as changing the version number in the .target file.

So this file is safe to modify? It says generated, so I was thinking some process need to run to create/update.

Bananeweizen:
All the changes I commit have been tested locally by
* changing the configuration
* clearing all markers
* running checkstyle again and verifying I have the same markers as before.

> Why isn't this also possible with the Checkstyle JAR?

That removal is probably also possible with the checkstyle jar. I did not yet do that, since I wanted to take a look at what else happens with that newly built checkstyle-all.jar and what https://github.com/checkstyle/eclipse-cs/pull/263 did instead. But it's on my todo list. :)

> So this file is safe to modify? It says generated, so I was thinking some process need to run to create/update.

Good catch, that line was a copy/paste error. I've taken time to update the readme and to explain the complete process in more details. It may look confusing on first glance, but it leads to a very simple structure for editing and generating files.

If you are curious, it looks like this now when opening the target file in a target editor: 
grafik

romani (approved):
I am approving, but @Calixte need to review this to be aware and may be new ideas come up how to simplify dependency on main library

net.sf.eclipsecs.target/readme.md
```diff
@@ -1,27 +1,54 @@
...
-Target files are hard to maintain manually. Therefore we use a custom DSL to
-maintain the target platform, and to derive the .target file.
+Target files are hard to maintain manually.
+Therefore we use a custom DSL to maintain most of the target platform, and to derive the `.target` file.
+However, that DSL based generation is not possible for the part of the target platform that shall include Maven coordinates.
+That's why the target platform is split over 2 files.
+The root target platform contains only the Maven coordinates and can be edited manually.
+It also includes a nested target file, which is generated from the target platform DSL.
+While that structure may look confusing at first glance, it simplifies the actual editing a lot,
+because you either only edit Maven coordinates or entries in the `.tpd`file.
```
Calixte, line 37:
```suggestion
because you either only edit Maven coordinates or entries in the `.tpd` file.
```

net.sf.eclipsecs.target/net.sf.eclipsecs.target.target
```diff
@@ -1,25 +1,30 @@
...
-      <unit id="org.junit.platform.runner" version="1.6.0.v20200203-2009"/>
-      <unit id="org.junit.platform.suite.api" version="1.6.0.v20200203-2009"/>
-      <repository location="https://download.eclipse.org/tools/orbit/downloads/drops/R20200831200620/repository"/>
-    </location>
-  </locations>
-  <targetJRE path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-11"/>
-</target>
+	<targetJRE path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-11"/>
+	<locations>
+		<location type="Target" uri="file:${project_loc:/net.sf.eclipsecs.core}/../net.sf.eclipsecs.target/net.sf.eclipsecs.partial.target"/>
```
Calixte, line 6: You use a "partial" target file here just for code organization / clarity?

Calixte (approved):

net.sf.eclipsecs.target/net.sf.eclipsecs.target.target
Bananeweizen: Yes. The Maven part of the target must be edited manually, and the remaining Eclipse part can be generated. However, you cannot have the generation and the manual editing in one target file, therefore the generated part is included as "partial". If you think there is a better name for more clarity ("generated" instead?), I'll happily accept every suggestion.

Calixte: Nope, partial sounds good to me. I'm just trying to make sure I understand.
We could take this a step further and make another partial for the maven libs, and the main target file would just be an aggregate of partials.

Bananeweizen: I would be fine with that, too. At work, I use exactly such a 3 target files setup. I wasn't sure if I confuse people even more if I go from 1 to 3 here. :)

PR #407, opened 2022-11-15
infra: gitignore temporary files of Tycho

Bananeweizen:
While the build is running, Tycho generates several temporary files outside of the Maven target directories. If developers refresh their staging view during the build, they may be confused by those temporary files. Therefore gitignore them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,9 @@ show.patch
 *.DS_Store
 net.sf.eclipsecs-updatesite/workspace
 *.zip
+
+# Tycho generates multiple files during the build.
+# They are excluded from the timestamp calculation by default, but still can confuse developers in the staging view.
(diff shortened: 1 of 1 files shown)


rnveach:
I assume these files are deleted when tycho is done? We run a git diff at the end of every travis run to look for modified files not committed and it hasn't failed.

Bananeweizen:
Yes, those files are deleted at the end. You only see them if you either terminate the maven build process somewhere in the middle or if you refresh the staging area in your git client in the middle of the build running. It's therefore really just for humans, not for the CI system. It's seldom, but I run into this every some weeks, typically when I use ctrl-c on a just started build, because I notice I have further editing to do.

romani (approved):
ok to merge

Issue #405, opened 2022-11-15
ExceptionInInitializer during workbench shutdown


Bananeweizen:
If the UI bundle of eclipsecs is not actively triggered during an Eclipse session, the following exception will be raised at shutdown. This is because the image descriptors for all the icons are initialized in a static block, which is only executed during shutdown of the workbench in that case, and some of the necessary objects are already gone.
The solution is to not use a static block there, and to convert the static image descriptors into static String members that are just keys to the image descriptor map. That way the creation of the image descriptors can be done lazy, but non-static.

The code was always wrong with the static block, but this exception can only appear since my recent PR changing the icons, because only since then this block accesses the workbench itself. I've only seen it during debugging sessions, I would not expect it in real life end user sessions.
```
org.osgi.framework.BundleException: Exception in net.sf.eclipsecs.ui.CheckstyleUIPlugin.stop() of bundle net.sf.eclipsecs.ui.
	at org.eclipse.osgi.internal.framework.BundleContextImpl.stop(BundleContextImpl.java:891)
	at org.eclipse.osgi.internal.framework.EquinoxBundle.stopWorker0(EquinoxBundle.java:1025)
	at org.eclipse.osgi.internal.framework.EquinoxBundle$EquinoxModule.stopWorker(EquinoxBundle.java:370)
	at org.eclipse.osgi.container.Module.doStop(Module.java:660)
	at org.eclipse.osgi.container.Module.stop(Module.java:521)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.decStartLevel(ModuleContainer.java:1886)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.doContainerStartLevel(ModuleContainer.java:1761)
	at org.eclipse.osgi.container.SystemModule.stopWorker(SystemModule.java:275)
	at org.eclipse.osgi.internal.framework.EquinoxBundle$SystemBundle$EquinoxSystemModule.stopWorker(EquinoxBundle.java:202)
	at org.eclipse.osgi.container.Module.doStop(Module.java:660)
	at org.eclipse.osgi.container.Module.stop(Module.java:521)
	at org.eclipse.osgi.container.SystemModule.stop(SystemModule.java:207)
	at org.eclipse.osgi.internal.framework.EquinoxBundle$SystemBundle$EquinoxSystemModule$1.run(EquinoxBundle.java:220)
	at java.base/java.lang.Thread.run(Thread.java:833)
[...]

Issue #388, opened 2022-11-12
mass clean up and bring back default config


Bananeweizen:
This issue is mainly to explain why many other things should be done first before bringing back the default checkstyle configuration. To recap, there are around 7000 issues with the default config. Fixing those should be as efficient as possible, and should not lead to contradictions with how the Eclipse IDE works where the eclipse-cs source is developed (because otherwise we may play ping-pong between Eclipse formatting something, Checkstyle complaining, manual fixing, and repeating the cycle all over).

[] Merge as many open PRs as possible. They would be in git conflicting state with all the reformatting changes coming later.
[] Enable automatic import organizing. Modify settings to match Checkstyle rules. Run organize imports on all sources.
[] Search/replace all the $NON-NLS markers in source code. Those are needed only for plugins which use bundle localization, which isn't the case for eclipse-cs.
[] Enable auto-format on save in Eclipse. Right now the formatter is not active at all, therefore newly written code doesn't fit the Checkstyle settings (like line length). Modify formatter settings to fit Checkstyle settings. Format all existing sources using the Eclipse formatter.
[] Enable cleanup on save in Eclipse. Enable as many options as possible. Run the cleanups. (Those cleanups work like automated refactorings. E.g. foo.equals("const") will automatically be reordered, thereby automatically fixing the Checkstyle issue.)
[] ... maybe more coming up along the way
[] enable standard config again, hoping the number should have gone down drastically

rnveach:
> Enable automatic import organizing
> Enable auto-format on save in Eclipse
> Enable cleanup on save in Eclipse

I just wanted to point out that while I do enable these for my own projects, these will only work if everyone is using Eclipse as their IDE. You can't expect all contributors to do this as I know for a fact some don't use Eclipse. This will cause extra differences in PRs between users who do use Eclipse and who don't. Some of the odd formatting you were talking about, may have been from someone not using Eclipse.

romani:
usage of Eclipse auto formatting is fine, but it should be in non-conflicting with checkstyle rules.

Bananeweizen:
> I just wanted to point out that while I do enable these for my own projects, these will only work if everyone is using Eclipse as their IDE

[...]

PR #410, opened 2022-11-17
Issue #405: exception during workbench shutdown

Bananeweizen:
Avoid a static block for creating the icon images, since we cannot really control when that static block is executed (it depends on the class loading time). Rather create the icons the first time they are accessed, since that is surely when the workbench is fully up and running.

Have tested configuration, execution, markers. Works as before, no more exception.

### DIFF
--- a/net.sf.eclipsecs.ui/src/net/sf/eclipsecs/ui/CheckstyleMarkerImageProvider.java
+++ b/net.sf.eclipsecs.ui/src/net/sf/eclipsecs/ui/CheckstyleMarkerImageProvider.java
@@ -29,7 +29,7 @@ import org.eclipse.ui.texteditor.IAnnotationImageProvider;
 
 /**
  * Image provider for Checkstyle markers.
- * 
+ *
  * @author Lars Ködderitzsch
  */
 public class CheckstyleMarkerImageProvider implements IAnnotationImageProvider {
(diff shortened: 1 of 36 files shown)


net.sf.eclipsecs.ui/src/net/sf/eclipsecs/ui/CheckstyleMarkerImageProvider.java
```diff
@@ -29,7 +29,7 @@
 
 /**
  * Image provider for Checkstyle markers.
- * 
+ *
```
romani, line 32: Sounds like auto formatting.
Is it possible to auto format whole project ones to avoid  such extra changes in future?

romani (commented):
I can not validate correctness of this PR.
Better if @Calixte  to look at it.

net.sf.eclipsecs.ui/src/net/sf/eclipsecs/ui/CheckstyleMarkerImageProvider.java
Bananeweizen: Yes, that's why I have it quite high in the bullet point list of #388. I also don't like reviews that contain whitespace and formatting changes.

PR #413, opened 2022-11-19
infra: remove java nature from non Java projects

Bananeweizen:
Eclipse plugins don't need to have Java source and nature. Therefore the docs and branding plugins can be without classpath and similar Java related settings.

Verified help and branding work as before. The most visible difference is that the 2 modified projects don't have the "J" overlay anymore for developers.


romani (approved):

Calixte:
Since this PR, I have either one of those two errors for the branding package:
- The plug-in's classpath library '.' does not have a corresponding source build entry
- No required execution environment has been set
Maven Update adds back the Java Nature, removes the first error, but adds the second one.

Bananeweizen:
@Calixte can you describe with more detail what you do when that comes back? Are you using the project explorer context menu Maven>Update Projects? If so, that is generally not required at all, AFAIK. The m2e integration in Eclipse is mostly relevant for projects that actually _need_ to execute some maven execution during the build, because they can't fully build otherwise. However, the Eclipse-CS plugin builds completely fine just using the Eclipse internal build mechanism, relying on the Eclipse Java compiler etc.
As far as I remember I've never used the Maven>Update Project menu on my Checkstyle workspace before.

PR #416, opened 2022-11-20
infra: update maven launch configs

Bananeweizen:
These launch configs can be used to run Maven from within Eclipse, without a separate shell. Update them in these aspects:
* have an additional launch config for "mvn verify"
* update all launch configs to the latest serialization format (including new preferences)
* disable parallel build for the update related launches, since those maven plugins are not thread-safe
* have already the next version number in the update-version launch config (this is fully working!)

There are no functional changes for the build, this only affects Eclipse developers working on the eclipse-cs code.


Bananeweizen:
for non Eclipse users, this is where those files appear: 
grafik

romani (approved):
Look like this update is not changing anything for users, only dev mode improvement

PR #419, opened 2022-11-27
Infra: Avoid repeated target file resolution

Bananeweizen:
Due to a bug in Tycho the path towards target platform files is not normalized. Therefore a/../target and b/../target are considered different files, even though they are in fact the same.

As a workaround use a path relative to the directory where the maven build was started in, which is identical for all maven modules. This is typically not recommended, because it would not work when building a single module only, but that doesn't work anyway in this project (due to the checkstyle maven config properties resolution).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -87,7 +87,9 @@
                 <version>${tycho-version}</version>
                 <configuration>
                     <target>
-                        <file>${basedir}/../net.sf.eclipsecs.target/net.sf.eclipsecs.target.target</file>
+                        <!-- Replace multi module dir by basedir after Tycho 3.0+ can be used.
+                        see https://github.com/eclipse-tycho/tycho/issues/1660 -->
+                        <file>${maven.multiModuleProjectDirectory}/net.sf.eclipsecs.target/net.sf.eclipsecs.target.target</file>
                     </target>
                     <resolver>p2</resolver>
                     <environments>


Bananeweizen:
For reviewers, the difference is that in this PR there is only one time the output for resolving the target file:
https://app.travis-ci.com/github/checkstyle/eclipse-cs/jobs/589554334#L486
In previous builds it is 8 times.

pom.xml
```diff
@@ -87,7 +87,9 @@
                 <version>${tycho-version}</version>
                 <configuration>
                     <target>
-                        <file>${basedir}/../net.sf.eclipsecs.target/net.sf.eclipsecs.target.target</file>
+                        <!-- TODO replace multi module dir by basedir after Tycho 3.0+
+                        https://github.com/eclipse-tycho/tycho/issues/1660 -->
```
romani, line 91: I know that most Checks disabled.
Todo is not allowed in final code.
Can you make it ` until https://github.com/eclipse-tycho/tycho/issues/1660` ?

romani (changes requested):
Items

pom.xml
Bananeweizen: I've reworded it, without TODO. "until URL" would be a bit confusing, since I have already fixed that bug in Tycho, but we will not be able to use the bugfix version anytime soon (remember the Java 17 dependency, when I tried to upgrade recently).

romani (approved):

pom.xml
romani: In this case we should create issue in our repo to migrate to jdk 17, and make comment to describe until it.

PR #423, opened 2022-12-05
Line endings

rnveach:
Discussions from at https://github.com/checkstyle/contribution/pull/717#issuecomment-1336195178

Checkstyle cannot enforce line endings as Git is the main authority, so this change ensure user's git will store files with respective line endings (Windows for this repo).

Please confirm the other commit's changes on files impacted by the gitattributes.


romani (approved):
But such files might be generated by eclipse

rnveach:
> But such files might be generated by eclipse

`Maven update check.launch` is an eclipse file, but I am not sure it matters what line ending it has. I am not really sure if I recognize `net.sf.eclipsecs.target.target`.

Bananeweizen (changes requested):
To me this doesn't look right. As far as I can tell, most files in the repository use Linux line endings. You can check by downloading an arbitrary file like https://raw.githubusercontent.com/checkstyle/eclipse-cs/master/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/CheckstylePluginPrefs.java directly from GitHub.

If you look at the same files in your Windows checkout, they should have Windows line endings (mine do), if you use Git for Windows and have not modified the config. git for windows sets `core.autocrlf=true` in the system config by default, which means the local files are modified by git on checkin/checkout.

In all the nested Eclipse projects the `org.eclipse.core.runtime.prefs` setting for `line.separator` is currently set to `\n`. I did that 2 years ago in 3f4584cbc503018a66f9b12349f277689b3050b8, which would support my thesis of the repo using mostly Linux line endings.

Before we go deeper with this change, what exactly is it meant to fix or avoid? The discussion in https://github.com/checkstyle/checkstyle/issues/893 already highlighted a lot of problems that can appear by being too strict in evaluating line endings. Right now the only change that seems _necessary_ to me is line 5 in the gitattributes (.sh files with LF), since such files would contain windows line endings on a default windows git checkout now.

rnveach:
Ok, it is interesting Travis is passing with this. I was getting a failure earlier on files it was modifying because I didn't have them as being ignored.

I forgot my own repo had autocrlf set on. I assumed we were using Windows line endings, but I did a fresh clone with git autoclrf off, and the files are indeed Linux line endings.

> Before we go deeper with this change, what exactly is it meant to fix or avoid?

Checkstyle cannot enforce the correct line endings in the repo. The mentioned PR above was one example where Checkstyle found this issue in one of it's sub-repos because of changes I made since I am a Windows user. I had checked in files that had Windows line endings instead of Linux.

Since this is a git issue, Checkstyle can't do anything about it. So the idea is to have gitattributes enforce the correct line endings. Yes, autocrlf can be used instead but this requires the user to manually add this before before submitting changes. Even a seasoned checkstyle user as me, forgot to do this in this one sub-repo. So instead of relying on the user to do this, and run into the possibility someone will forget, this change is to ensure it never happens.

> Right now the only change that seems necessary to me is line 5 in the gitattributes (.sh files with LF), since such files would contain windows line endings on a default windows git checkout now.

I basically inverted this file from main repo since I thought this repo was using Windows line endings, but that is wrong.

`* eol=crlf` should be swapped to `LF` as discussed.

````# Linux script files always have LF eol
*.sh eol=lf````
Should be swapped to Batch files which need to be `CRLF`.

> The discussion in https://github.com/checkstyle/checkstyle/issues/893 already highlighted a lot of problems that can appear by being too strict in evaluating line endings.

I am trying to skim through the discussions to refresh my memory, but is there any issues with Windows use `LF` also? Gitattributes will enforce the line ending we specify. I have no issues in other repos, but I am not living in eclipsecs as the others.

We can specify file formats for correct line endings, so batch/shell is not an issue.

rnveach:
I confirmed on my local with autocrlf off that `net.sf.eclipsecs.target.target` and `Maven update check.launch` does have Windows line endings.

Bananeweizen (approved):

rnveach:
Normally I don't install a Git client. I use to go solely by egit in Eclipse. I dunno if they changed like you suggested, but everything for me required me to manually turn on autocrlf. I never went anywhere global and turned it off. I did have 1 company computer automatically turn it on, but I assume that was predefined setting by the company's installer and not the new default.

Bananeweizen:
@rnveach Hah, that's one more problem then. :) The core egit developers assume that you have an additional native git client installed. I can make that statement since I'm also egit committer (but not really active anymore). I have been very surprised by that in the beginning myself, assuming that egit is fully sufficient when using eclipse. However, jgit and egit are just not fully feature compatible with native git, and you may sometimes run into problems that would not exist with native git installed.

PR #425, opened 2022-12-11
don't use input text for filter hints

Bananeweizen:
JFace text controls have a separate message property to give hints for the expected input into a text field. Programmatically setting the actual control text instead is a bad idea since it can lead to the user actually editing that hint.

before the change its possible to edit the hint, since it's the actual widget text:
grafik

after the change the widget itself shows the hint only if the input is empty:
grafik

### DIFF
--- a/net.sf.eclipsecs.ui/src/net/sf/eclipsecs/ui/config/CheckConfigurationConfigureDialog.java
+++ b/net.sf.eclipsecs.ui/src/net/sf/eclipsecs/ui/config/CheckConfigurationConfigureDialog.java
@@ -72,8 +72,6 @@ import org.eclipse.osgi.util.NLS;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.browser.Browser;
 import org.eclipse.swt.custom.SashForm;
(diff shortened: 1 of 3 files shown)


romani (approved):
We recommend for commits that has more details in PR description than in commit message, reference PR number in commit.

`#425: dont .....`
To let future engineer to look more details if some changes become problematic

Bananeweizen:
> We recommend for commits that has more details in PR description than in commit message, reference PR number in commit.

I know, but the commit message has the same details (except for the images) and there is no issue.

And generally speaking, IMO the main checkstyle project has a really bad rule for commit messages by limiting their length and thereby enforcing people to document the change in the issue tracker instead. Basically every other project I know requires the commit message to describe both what and why something changed, such that the git history itself is sufficient for understanding why commits were created. Issue trackers are regularly replaced by new software, or someone might work offline. But the git history definitely is available at the developers machine, and that's why the reason for a change should be fully in the commit message. Referencing an issue is also a good practice, but moving the commit message details to the issue tracker is not. In fact, I have given up 2 or 3 trivial contributions to the main checkstyle project when they failed the build because of that.

PR #426, opened 2022-12-11
fix #375: don't block UI when loading metadata

### LINKED ISSUE #375: creating new config blocks main thread
Creating a new configuration from the preference page leads to blocking the main thread for several seconds. That operation should be run via a UI job to avoid "application not responding".

Bananeweizen:
Use a progress dialog with indeterminate progress while loading the metadata. That way the UI thread is not blocked, and Windows does not show "application not responding". The progress dialog automatically closes after metadata loading has finished.

grafik


romani (approved):
I surprised that this is not instant

Calixte:
> I surprised that this is not instant

Me too. It's nearly instantaneous on my machine.

Bananeweizen:
> It's nearly instantaneous on my machine.

That's when it has been loaded by another action (like running scan) before. If you just start Eclipse without any build running, and then go to the configuration, you will see quite a delay (some seconds in my case), but only the first time.

rnveach:
This also shows up in the Error Log view. I assume this fix has not been released yet.

ui freeze

PR #429, opened 2022-12-11
Avoid wrong checkstyle library being unpacked

Bananeweizen:
If multiple zip files are available in the local directory (from previous builds of older releases), the previous code might choose a wrong outdated version. Just use the known file name from the download to avoid that.


Bananeweizen:
I noticed that my local build would use the outdated 10.3.3.zip because of this (even though everything else was 10.4).

PR #431, opened 2022-12-12
fix #430: set unused button invisible

### LINKED ISSUE #430: invisible widget
When using tab to cycle through the buttons in the configuration editor, there is an invisible component between "Remove" and "Export" which actually is focused during the tab cycle, but cannot be seen by the user.

Bananeweizen:
https://github.com/checkstyle/eclipse-cs/issues/430

Previously that button was just not layouted in the container widget, but it was actually usable. Now set the widget to invisible, thereby removing it from the tab key cycling and from the button click listener.


rnveach:
@Bananeweizen When you return, please rebase on the latest master and let us know if you tested this as working.

Bananeweizen:
Of course I tested this. I never push anything that has not been working locally. And stuff like this can't be shown with a screenshot...

PR #434, opened 2022-12-13
Improve blueprint selection dialog

Bananeweizen:
Use a more specialized dialog to select the soure project for copying a checkstyle configuration. This brings several benefits:

* allows filtering the projects via search mask
* automatically sets selection to first project in list
* filter out closed projects (or otherwise inaccessible projects)
* enable OK button only if there is a selection

The dialog still looks very similar to how it looked before:
grafik


rnveach:
@Bananeweizen There is a conflict now.

Calixte (approved):

romani:
@Bananeweizen , please resolve conflict and let merge this PR.

Bananeweizen:
rebased and tested again

Issue #365, opened 2022-10-28
Need new maintainer?


vtintillier:
Hello @romani ,

as mentioned in #364 already, @Calixte had no GitHub activity since May 1st with https://github.com/checkstyle/eclipse-cs-update-site/commit/515038931f10fb9869625583a68e80337c0f1ef9

Maybe that would be a good time for new maintainers to be added to the checkstyle/eclipse-cs and checkstyle/eclipse-cs-update-site repositories to keep this project alive?

That's something I could do, and I also see e.g. @Bananeweizen who has a few opened PR, who might be interested as well?

Thanks for your help

romani:
> That's something I could do

we always welcome new people to project. Please try to fix few issues to understand how project works.

I connected with @Calixte  by email, he responded (7 days ago) that he might have time to do releases this week.
But your involvement is always welcome and as you get some experience in the project we are ok to grant you permissions to release project. Please join a team by starting of contribution.

Bananeweizen:
While I don't have much time myself for contributions, I must admit it _feels_ worse since the Eclipse plugin joined the main checkstyle project. The reviewers of the main checkstyle project typically don't know the Eclipse API, therefore they can't really judge new PRs, which then tend to be open very long (and I don't want to blame the people reviewing the changes, I'm rather looking at the process as such). In the former eclipse-cs setup, I would just merge all my open PRs with when no reviews came in after some days. If we could somehow improve that process, contributions to the Eclipse part would become both easier (I previously could not finish other improvements that depend on open PRs) and more attractive to new contributors.
Any thoughts by others on that?

rnveach:
I can only speak for myself.

I agree, I don't know the Eclipse API and can only really judge code style and things that stand out to us as questionable and in some rare cases, wrong. I have tried to do my own PRs but then some were reverted as wrong and even then I never got a clear picture of why it was wrong.

I am ok with changes, but I am unclear on what you are suggesting? Based on your comment, you want to merge your own works and take control of old PRs? I see no issue with the later, checkstyle maintainers do this too when things are becoming abandoned. For merging, I have no been active in this repo because of my lack of knowledge. But if it makes things to easier to just merge faster, then if you are ready for something to be merged, I would say ping and assign me and I will give a quick glance and approve/merge if nothing stands out to move along the process if that is what we want.

romani:
>must admit it feels worse since the Eclipse plugin joined the main checkstyle project.

[...]

PR #435, opened 2022-12-17
Bulk fix Checkstyle violations with many occurrences

Bananeweizen:
Almost all of the changes have been done in semi automated fashion (e.g. Rename refactoring tool, search replace for removing Javadocs etc.). There are no functional changes to the code.

This fixes the issues with the most occurrences that would appear when re-enabling the currently disabled checks. Applying these fixes now is easier, because Checkstyle will not (yet) complain about the remaining issue in the modified files.


romani:
27 more violations.
https://app.travis-ci.com/github/checkstyle/eclipse-cs/jobs/591481693#L522

If they are more complicated I recommend to suppress them temporarily until next PRs.

Bananeweizen:
No, this doesn't introduce any new violations. This change rather fixes around 500 violations. The build just reports 27 of the remaining 17.000 violations in touched files, because the patch filter setting doesn't know that the checks are generally disabled. And since those 27 violations are part of another bulk fix (fixing violations by type, not by file), it doesn't make sense to insist on fixing them separately now. I have another change to fix around 16.000 violations coming up, so please just accept this change.

rnveach:
> because the patch filter setting doesn't know that the checks are generally disabled

Patch filter only works on checks that are enabled. It can't see disabled ones. Since CI is getting violations, then it stands that those checks are enabled.

net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/projectconfig/PluginFilters.java
```diff
@@ -106,8 +106,8 @@ public final class PluginFilters {
         filter.setFilterData(data);
 
         filters.add(filter);
-      } catch (Exception e) {
-        CheckstyleLog.log(e);
+      } catch (Exception ex) {
```
rnveach, line 109: > PluginFilters.java:[109,9] (extension) IllegalCatchExtended: Catching 'Exception' is not allowed.

Enabled at:
https://github.com/checkstyle/eclipse-cs/blob/5b295acc79fd1bc37e38d9a2f06c01667c57ed29/config/checkstyle_sevntu_checks.xml#L93-L95

rnveach (commented):
Just showing one example that looks valid that it is being reported since the line was changed.

Bananeweizen:
@rnveach That check checks for the _type_ of exception, which hasn't changed in that line. So the same violation was there before, but it was not reported, because that check is disabled in checkstyle_checks. Maybe the patch filter uses checkstyle_sevntu.xml instead, I don't know. And the exact technical difference is not even relevant.

We have disabled a big amount of checks in the checkstyle_checks.xml, because it's almost impossible to commit a PR that fixes violations, because CI will always complain about the _remaining_ some thousand violations. In the end we would need a single PR to fix all 18.000 violations, if the checks were still enabled.
And similar to how we ignore almost all violations as such _right now_, I want to ignore them when being raised from the patch filter _for these bulk cleanups_. Just for efficiency. I can easily fix 100 identical violations in a short amount of time, but I cannot fix another 20 random different issues raised by the patch filter afterwards, because it takes much more time to investigate each type of violation first. I thought I had already explained that in the issue for "bring back checkstyle config".

This is one more example for what I mentioned in https://github.com/checkstyle/eclipse-cs/issues/365#issuecomment-1296249270. I know very well that I introduced no additional violations in this PR and in the setup before joining the checkstyle project I would have merged this PR meanwhile, and fixed another some hundred violations in the time I spent for this discussion. I'm meanwhile back at a mood where I think about no longer contributing, because for me personally the benefits from the actual software improvements are not worth all the stress induced by such back-and-forth discussions.

rnveach:
> because that check is disabled in checkstyle_checks

Please look at the XML file I referenced and the travis logs. The configuration for the violation is coming from [checkstyle_sevntu_checks.xml](https://github.com/checkstyle/eclipse-cs/blob/5b295acc79fd1bc37e38d9a2f06c01667c57ed29/config/checkstyle_sevntu_checks.xml#L93-L95) and is not disabled. This repo has 2 configurations set up, separated by Checkstyle only checks an Sevntu only checks. Like main repo, we don't have one config combine both.

> Maybe the patch filter uses checkstyle_sevntu.xml instead

It is the other way around. Checkstyle is the one that uses patch filter as it is a suppression filter added to the configuration file. They can be seen at 
https://github.com/checkstyle/eclipse-cs/blob/master/config/checkstyle_sevntu_checks.xml#L15 and https://github.com/checkstyle/eclipse-cs/blob/master/config/checkstyle_sevntu_checks.xml#L21

> it's almost impossible to commit a PR that fixes violations, because CI will always complain about the remaining some thousand violations
> we would need a single PR to fix all 18.000 violations, if the checks were still enabled

This is because the patch filters were removed  from `checkstyle_checks.xml`. If it was re-added, it would be a lot easier to maintain like sevntu's config. You would only get violations on the lines that were changed as seen in this PR.

> by such back-and-forth discussions.

I am not back and forthing. I am just pointing out how patch filter works and the misunderstandings the behavior being seen. I have not said somethng needs to be fixed. I am leaving this for others on how to handle.

romani:
>  so please just accept this change.

I am always ok to improve step by step, but CI should be green on each step. Red CI cause too much confusions for others and mess.

you can keep checkstyle disabled to archive this.
So you activate checkstyle, fix something, disable it back.

rnveach:
@romani Since this PR has been sitting for a while, I rebased it and added a new commit which should hopefully resolve CI.

I mostly added suppressions for the majority, as I am not comfortable resolving them yet and will need to be reviewed later.

rnveach:
CI should pass now.

rnveach (approved):

romani (approved):
Ok to merge

PR #436, opened 2022-12-26
config: upgrade sevntu to 1.44.1

rnveach:


romani:
This version has no functional change

Issue #456, opened 2023-01-29
Conflict of Files processed by Checkstyle in Repo during Maven run and Eclipse-CS Plugin Run


rnveach:
Noticed after merger of 74e9af6437cae440bd01e60467599a771549ea94 and working on my Eclipse-CS local to properly compile.

When I open a clean version of the repo, I now get the following violation:
````
Description	Resource	Path	Location	Type
File does not end with a newline.	messages.properties	/net.sf.eclipsecs.sample/src/net/sf/eclipsecs/sample/checks	line 1	Checkstyle Problem
````

However `mvn verify` passes.

I expected maven and eclipse-cs plugin to be synched for this repo and produce the same results.

rnveach:
I think there is a few issues that will lead to conflicts.

First, not every sub-module has a `.checkstyle` file. This is what tells eclipse-cs plugin to run checkstyle on the project. When running through maven, checkstyle is run on every module as seen from this cut out snipper:
````
[INFO] --- maven-checkstyle-plugin:3.2.0:check (checkstyle-check) @ net.sf.eclipsecs.parent ---
[INFO] --- maven-checkstyle-plugin:3.2.0:check (sevntu-checkstyle-check) @ net.sf.eclipsecs.parent ---
[INFO] --- maven-checkstyle-plugin:3.2.0:check (checkstyle-check) @ net.sf.eclipsecs.branding ---
[INFO] --- maven-checkstyle-plugin:3.2.0:check (checkstyle-check) @ net.sf.eclipsecs.doc ---
[INFO] --- maven-checkstyle-plugin:3.2.0:check (sevntu-checkstyle-check) @ net.sf.eclipsecs.branding ---
[INFO] --- maven-checkstyle-plugin:3.2.0:check (sevntu-checkstyle-check) @ net.sf.eclipsecs.doc ---
[INFO] --- maven-checkstyle-plugin:3.2.0:check (checkstyle-check) @ net.sf.eclipsecs.checkstyle ---
[INFO] --- maven-checkstyle-plugin:3.2.0:check (sevntu-checkstyle-check) @ net.sf.eclipsecs.checkstyle ---
[INFO] --- maven-checkstyle-plugin:3.2.0:check (checkstyle-check) @ net.sf.eclipsecs.core ---
[INFO] --- maven-checkstyle-plugin:3.2.0:check (sevntu-checkstyle-check) @ net.sf.eclipsecs.core ---
[...]

PR #457, opened 2023-01-29
 Issue #456: synchronizes eclipse-cs and maven 

rnveach:
Issue #456

https://github.com/checkstyle/eclipse-cs/issues/456#issuecomment-1407569510
This resolves everything except 
> maven-checkstyle is set to look for folder structures like src\main\resources

I wasn't sure if we wanted to make such a change in eclipse. I look at egit/jgit, which is another eclipse plugin, and it doesn't do this. I am not sure if eclipse projects prefer to keep this way, unlike normal maven projects.


net.sf.eclipsecs.checkstyle/metadata/com/puppycrawl/tools/checkstyle/checkstyle-metadata_1_0.dtd
```diff
@@ -8,138 +8,138 @@
 -->
```
rnveach, line 8: @Calixte What is the purpose of https://github.com/checkstyle/eclipse-cs/tree/master/docs and how is it updated?

I noticed it has the same DTD files at https://github.com/checkstyle/eclipse-cs/tree/master/docs/dtds that I updated in this PR, but nothing required those files to be updated as well.

Calixte: The `/docs` directory is the directory configured to be served by GitHub Pages at https://checkstyle.org/eclipse-cs/
It is automatically generated during the release when running `mvn package` on `net.sf.eclipsecs.doc`.
I believe the two are decoupled so the source doc `net.sf.eclipsecs.doc` can be used to generate the website, but also the Eclipse Help content.
See those lines in the release script: https://github.com/checkstyle/eclipse-cs/blob/master/release.sh#L70-L75

rnveach: So basically `/docs` is the suppose to be the last release's documentation (10.4.0) as we are in snapshot for the next release. Eclipse-CS doesn't use another repo for the hosted site like main repo does at https://github.com/checkstyle/checkstyle.github.io ?

So basically https://checkstyle.org/eclipse-cs/ is a mirror of http://checkstyle.github.io/eclipse-cs/ which points to https://github.com/checkstyle/eclipse-cs/tree/master/docs/ in virtually real time?

Calixte: You are correct.

rnveach: Thanks @Calixte

Calixte: > nothing required those files to be updated as well

Actually, those fail when I build now.

Issue #451, opened 2023-01-27
Synchronize Checkstyle Config from Main Repo


rnveach:
Start the process of synching Checkstyle's config.

Some may be tied to https://github.com/checkstyle/eclipse-cs/issues/388 and thus may not be able to be completed.

PR #460, opened 2023-01-31
Issue #451: fix checkstyle violations

Calixte:
We're missing translations for https://github.com/checkstyle/eclipse-cs/blob/master/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/messages.properties so we can't require translations.


rnveach:
This was a conflict between 24e563882ab4272c769e4b055afb8684ddad51de and b860784aa58221c3eb5c4521d3ed1aa265de5578 that were merged simultaneously in separate PRs. This only came up after both were merged into 1 branch.

PR #463, opened 2023-02-02
config: removes custom arch to avoid credit usage

rnveach:
Travis can only use arch `ppc64le` otherwise we get charged. Travis will shutdown when we use up all our credits.

We won't know if this will continue working once travis returns, but we need this removed regardless.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,6 @@ jobs:
 
     # JDK 17 (most recent Java LTS version)
     - jdk: openjdk17
-      arch: amd64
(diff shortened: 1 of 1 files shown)


romani (approved):

romani:
Amd was not free arch.  Now I remember having problems to run eclipse stuff on free tier of Travis.

rnveach:
If travis can't support us, GH workflow will at https://github.com/checkstyle/eclipse-cs/pull/464 .

Issue #467, opened 2023-02-03
The currently displayed page contains invalid values dialog


rnveach:
While working on https://github.com/checkstyle/eclipse-cs/issues/454 ,

I guess I created an invalid configuration file. I got the following in the Problem View,

| Description | Resource | Path | Location | Type |
| --- | --- | --- | --- | --- |
| Checkstyle execution failed due to an internal error. Please check the error log for details... | net.sf.eclipsecs.checkstyle | | Unknown | Checkstyle Problem |

When I right clicked on the problem and did properties I got this error dialog which made no sense:
````
Could Not Accept Changes

The currently displayed page contains invalid values.
````
![The currently displayed page contains invalid values](https://user-images.githubusercontent.com/5427943/216506126-3131d19a-063a-403a-be34-b7aff40e9e2c.png)

I assume that dialog should not appear, and even still, I think there should be some way to communicate the error that Checkstyle provides without going to the Error view. I see the Error view as more a last resort when everything else has failed and it is some exception with the plugin.

The error I was getting turned out to be: `Caused by: com.puppycrawl.tools.checkstyle.api.CheckstyleException: Property ${checkstyle.regexp.old.header.file} has not been set`

While this is an error in Checkstyle, it should be more of an expected exception from the plugin that should be handled gracefully somehow.

Bananeweizen:
Please post _everything_ from the stacktrace of the error in the error log. That error is not really relevant, but rather where it occurred in the plugin code.

rnveach:
Error log:

````
Plugin: org.eclipse.jface

Message: Problems occurred when invoking code from plug-in: "org.eclipse.jface".

Exception Stack Trace:

java.lang.NullPointerException: Cannot invoke "net.sf.eclipsecs.core.config.meta.RuleMetadata.getGroup()" because "metaData" is null
	at net.sf.eclipsecs.ui.properties.marker.MarkerPropertyPage.createContents(MarkerPropertyPage.java:62)
	at org.eclipse.jface.preference.PreferencePage.createControl(PreferencePage.java:244)
	at org.eclipse.jface.preference.PreferenceDialog.createPageControl(PreferenceDialog.java:1433)
	at org.eclipse.jface.preference.PreferenceDialog$8.run(PreferenceDialog.java:1196)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:45)
	at org.eclipse.jface.util.SafeRunnable.run(SafeRunnable.java:174)
[...]

PR #472, opened 2023-02-05
Issue 467: Avoid property page for runtime error marker

Bananeweizen:
Issue #467

Tested.

### DIFF
--- a/net.sf.eclipsecs.ui/src/net/sf/eclipsecs/ui/properties/marker/CheckstyleMarkerPropertyTester.java
+++ b/net.sf.eclipsecs.ui/src/net/sf/eclipsecs/ui/properties/marker/CheckstyleMarkerPropertyTester.java
@@ -19,7 +19,12 @@ public class CheckstyleMarkerPropertyTester extends PropertyTester {
     }
     IMarker marker = (IMarker) receiver;
     try {
-      return CheckstyleMarker.MARKER_ID.equals(marker.getType());
+      if (!CheckstyleMarker.MARKER_ID.equals(marker.getType())) {
+        return false;
+      }
+      // avoid property page for markers that show runtime errors instead of violations
+      Object module = marker.getAttribute(CheckstyleMarker.MODULE_NAME);
+      return module instanceof String && !((String) module).isBlank();
     } catch (CoreException ex) {
       CheckstyleLog.log(ex);
     }


rnveach (changes requested):
Please provide a screenshot of the new dialog displayed.

Edit: Also please clarify what "Tested" means. Did you just try to re-open the properties dialog from the issue case? Do valid violations still work? etc...

rnveach:
Travis is failing, but it looks like it is a difference from another change. GH workflow is missing diff at the end of run.

rnveach (changes requested):
As a secondary note, the code at https://github.com/checkstyle/eclipse-cs/blob/84f1b6103c96c01928bc50470d57e661bbea2a21/net.sf.eclipsecs.ui/src/net/sf/eclipsecs/ui/properties/marker/MarkerPropertyPage.java#L62 still remains the same.

The original exception occurred because we were looking for metadata on a module named the same as a blank string which obviously didn't exist.

Since the underlying code wasn't changed, we can still get an exception when metadata doesn't contain a module that produced a violation. Sevntu is an example of a 3rd party that has modules which do not use metadata. Sevntu doesn't use metadata because https://github.com/checkstyle/eclipse-cs/issues/316 is still un-fixed.

Even if the change to `CheckstyleMarkerPropertyTester.java` is valid, we still need to fix the underlying code. Maybe have similar code to what was there before metadata integration.

rnveach:
It seems to me we need to start examining the possibility of adding more automated tests to Eclipse-CS. I am just learning Eclipse plugin development, so I have no clue on what automated testing would look like.

Bananeweizen:
"Tested" means "I have done the same as in every other PR and project I work on, both positive and negative testing. But in other projects people don't ask whether that was tested, and that's why I don't need to mention it there."

The properties dialog no longer contains the Checkstyle specific page, in case of the runtime error marker that you reported this for. The ...Tester class is the one that decides whether that page is contributed for the object to show properties for. That's why all the code of the property page itself is not really relevant for this bug. More checks for missing metadata might be good, but really not relevant for this bugfix.

Your review comments come across in a way that sounds (for me) like this bugfix is completely questionable, and like I cannot be trusted to know what I'm doing (although I have contributed most changes to the project after the initial author, and although I'm an Eclipse platform committer). Would you mind rather asking about how it works instead of questioning it? That would make my side of it all more enjoyable.

rnveach:
I ask about what I don't know about or looking to understand or verify, like "What does the new result look like". I find asking direct questions for my missing gaps helps me more than being broad and possibly having to re-ask again when something specific I look for isn't mentioned. I am sure other projects ask less because they know the project better than I know eclipse-cs and eclipse in general. I ask less in main repo since I know it so well, but also all our automated tests and reports help reduce questions. You know how to test the application, but from all my [discussions](https://github.com/checkstyle/eclipse-cs/discussions) I am still learning eclipse specific things and wanted to know specifics of what does "tested" mean so I know and can assist if there is maybe something else I think is missing (like https://github.com/checkstyle/eclipse-cs/pull/472#pullrequestreview-1284330791). Also, I find providing a bit more documentation helps when issues become stale or a new issue comes up and want to look at an old issue/PR to see where things went wrong and what was missed. 5 months down, saying less doesn't help refresh your memory and it is easy to forget.

Really, most of my questions were only asked because I couldn't install the changes myself to test because of https://github.com/checkstyle/eclipse-cs/discussions/473 . Otherwise I would have looking at things myself to be a bit more informed and would have helped provided things before asking blind things.

When reproducing the issue and opening the properties, I am now given the following dialog with "Checkstyle" completely gone:
image

I am having issues producing normal violations and am trying to look into my issue. I don't see it being related to this PR.

Bananeweizen:
> When reproducing the issue and opening the properties, I am now given the following dialog with "Checkstyle" completely gone: 

That's the intention of the bugfix. The properties dialog is always the same for any object in Eclipse, just the number of pages being shown is different and depends on what plugins decide to contribute. Compare to the properties of a file, a project etc. The Checkstyle page is intended to be contributed for markers with violations, so it's gone for the runtime error marker now as it was never intended to be shown for your marker.

romani:
@rnveach , are you ok with answer on your question ?

rnveach (approved):

PR #477, opened 2023-02-06
minor: locks down method access in MetadataFactory

rnveach:
Methods are never accessed outside the class and do not need to be public.


Bananeweizen:
If you want to do that large scale, then running UCDetector would help in finding (and quick-fixing) all changeable visibility modifiers instead of manually looking at them. However, all of those findings would have to be taken carefully, since that tool doesn't know anything about conventions, where objects in Eclipse are used via reflection, by naming conventions, creation of adapters and similar things that may look like "not needed publicly", but then breaks at runtime.

Bananeweizen (approved):

romani (approved):

romani:
this is minor update, not functional, we can merge without @Calixte .

PR #478, opened 2023-02-06
Fix #475: include transitive dependencies

### LINKED ISSUE #475: How to create installable Eclipse-CS snapshot plugin artifacts?
### Discussed in https://github.com/checkstyle/eclipse-cs/discussions/473

<div type='discussions-op-text'>

<sup>Originally posted by **rnveach** February  5, 2023</sup>
I run `mvn clean package` on master and get a successful build. I try to go to Eclipse and install the plugin from `eclipse-cs\net.sf.eclipsecs-updatesite\target\repository` but it gets the following error:

````
Cannot complete the install because one or more required items could not be found.
  Software being installed: Eclipse Checkstyle Plug-in 10.4.0.202302051801 (net.sf.eclipsecs.feature.group 10.4.0.202302051801)
  Missing requirement: Eclipse Checkstyle 10.4.0.202302051801 (net.sf.eclipsecs.core 10.4.0.202302051801) requires 'java.package; org.dom4j 2.1.3' but it could not be found
````

When comparing my packaged version versus the release, I notice that the core JAR is missing the 3rd party JARs (dom4j, snakeyml, javaassist). I assume this is why Eclipse fails to install, but I am not seeing what is missing to include them.

I also noticed my doc JAR is 1000x larger than the release version, but I assume this is unrelated.

This is the same with `net.sf.eclipsecs-updatesite/target/net.sf.eclipsecs-updatesite_10.4.0.202302051801.zip`.

I have no issues compiling the project in Eclipse (no errors).</div>

===========

It appears 32d604929ad7f24488cd175e2f587b49986d5056 broke creating usable Eclipse-CS plugin artifacts with Eclipse-CS' own package instructions.

https://github.com/checkstyle/eclipse-cs/blob/563ce31909a22dfd581d58498c5d79f26bd6a84b/README.md?plain=1#L36-L38

Creating an artifact from 7e02d4eb28eaf8b379d4215556f5604fb382970d (the commit directly before 32d604929ad7f24488cd175e2f587b49986d5056) produces an plugin artifact with the necessary JARs in the lib folder. I tested and was able to install the artifact produced without issue.

Starting with 32d604929ad7f24488cd175e2f587b49986d5056, the necessary JARs are missing and I can no longer install what was produced.

I am unsure if this an instructional issue or if it is a bigger issue.

Bananeweizen:
Fix #475

This way Tycho automatically includes direct _and_ transitive dependencies in the update site. That is more than just the missing libraries, it includes every plugin needed when installing Eclipse-CS (that is, down to the actual Eclipse workbench and similar plugins).

The amount of data used during an installation does not change. The main change is that during installation all _needed_ plugins are found in the Eclipse-CS update site _as such_ while previously other known Eclipse repositories would be contacted to look for the missing transitively required bundles. This setting should have been used before, as it is generally good practice to have self-contained update sites.

Users can now also uncheck the checkbox "Contact all update sites to find required software", which will shorten the time drastically.

Below is just the beginning of the plugin list now contained in the update site:
grafik

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -289,6 +289,15 @@
                         </execution>
                     </executions>
                 </plugin>
+                <plugin>
+                    <groupId>org.eclipse.tycho</groupId>
+                    <artifactId>tycho-p2-repository-plugin</artifactId>
+                    <version>${tycho-version}</version>
+                    <configuration>
+                        <!-- make the update site self-contained by including all transitive dependencies -->
+                        <includeAllDependencies>true</includeAllDependencies>
+                    </configuration>
+                </plugin>
                 <!-- Disable default deployer. -->
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>


rnveach (approved):
I am able to install the plugin it creates now.

I noticed after an install, before restarting, I got:
````
eclipse.buildId=4.24.0.I20220607-0700
java.version=17.0.3
java.vendor=Eclipse Adoptium
BootLoader constants: OS=win32, ARCH=x86_64, WS=win32, NL=en_US
Framework arguments:  -product org.eclipse.epp.package.rcp.product
Command-line arguments:  -os win32 -ws win32 -arch x86_64 -product org.eclipse.epp.package.rcp.product

org.eclipse.equinox.frameworkadmin.equinox
Warning
Mon Feb 06 14:29:23 EST 2023
Failed creating shared configuration url for null.
````
With no information. Do not know if it is related.

I noticed with this change there are more unsigned ¿bundles? to approve during the install. I believe this makes sense and is not really an issue.
image

I did notice though that the installable artifact ballooned 3x in size from 35 megs to 107 megs.

I understand this is because it is now complete without needing to contact the update site. I still question if it is fully necessary.

I leave this for @Calixte to decide.

Bananeweizen:
There would be an alternative to have a smaller size, but that is error prone. Since the manifest information about bundle dependencies is only evaluated at startup time of the OSGI application (Eclipse), it has absolutely no meaning for _installation_ time (from the update site). That's why features have to kind of "redeclare and group" all the same dependencies to say which feature contains which bundle, depends on what other feature, requires what other feature/bundle to exist before its own installation etc.

Testing the completeness of the feature content in an automated fashion is really hard (I don't even do that for commercially created Eclipse installations), that's why this easily breaks after developers changed or updated plugin dependencies. Therefore enabling the automated transitive inclusion will create much less future issues about "Cannot install because of missing dependency x.y.z".

Issue #465, opened 2023-02-03
Eclipse-CS does not refresh Checkstyle's External Resource list


rnveach:
While working on https://github.com/checkstyle/eclipse-cs/issues/454 ,

I noticed when I changed the regular expression header file and did a clean build to re-run Checkstyle, Eclipse-CS kept using the old header file not matter what I did to try to get it to see the new one.

Checkstyle did not recognize the new header file until I re-saved the Configuration file, even though it had no more changes at that point.

It seems to me Eclipse-CS is caching something in memory to avoid having Checkstyle recreate the config in memory for each run.

rnveach:
The issue is the way Eclipse-CS runs Checkstyle.

`Auditor` is basically the Checkstyle runner and the [Checker instance](https://github.com/checkstyle/eclipse-cs/blob/dd602eb158f940803f702bb6c19d9d5ed206b9a1/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/builder/Auditor.java#L139) it creates is cached throughout the life of the Checker run.

Checker assumes the instance is always re-created on a run, so Modules are never set to reload files they have loaded after the first time. There is no check for an external resource has changed. There is just loading and destroying.
Header Check Example: https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/checks/header/AbstractHeaderCheck.java#L194-L197

For fixing this, it seems the only way may be to remove the cache on Checker.

I doubt Checkstyle will want to build a reloading lifestyle that it currently doesn't support and also technically can't use in it's own run lifecycle. While checker manages the external resources from the configuration and modules, it only does so to pass to the internal cache it manages when it is enabled.

Maybe Eclipse-CS could make use of Checkstyle's caching module and use that to trigger a reset, but this isn't supported by RootModule's in general (which Eclipse-CS also doesn't support) and even if it was, it only supports giving a file and does not provide any feedback that Eclipse-CS could really use.

Bananeweizen:
use ctrl-3, purge... to get rid of the eclipsecs cache.

PR #488, opened 2023-04-02
WIP: use checkstyle core cachefile

Bananeweizen:
This is a proof of concept only and needs further discussion.
When running checkstyle via maven plugin, it's really easy to use the cacheFile mechanism for more speed. In the eclipse plugin, there is no caching enabled for checkstyle core (the plugin only caches some of its own configuration files and similar). This leads to very long runtime in case of clean builds (because all files will be scanned again then, independent of whether there are actual changes). I have some workspace where checkstyle consumes 20 minutes of CPU time on clean builds.
This change enables the cacheFile mechanism in checkstyle core. It uses a single cachefile for all checkers, which is stored in the ".metadata" folder belonging to a workspace. That way the cachefile is not visible in the workspace itself, and every workspace has its own persistent cachefile, without using any global temp directories or similar.
The cache is not configurable and always enabled. It can be invalidated with the already existing "Purge checkstyle caches" action (hit Ctrl-3, start typing "purge").
The effect of the cache can be seen best when using the context menu on a large project to invoke checkstyle on it. Without the cache that always takes the same time, with a visible job in the progress view. With the cache the first invocation takes that same long time, every additional invocation on an already checked project is super-fast (in fact, I don't see a job anymore in the progress view in most cases because the job is finished more quickly then the progress view updates).

Open issues:
* The existing lifecycle of the cachefile in checkers doesn't really fit for the plugin, therefore it currently calls the persist method via reflection. This method should be made public API in the checkstyle project. Right now it's only called from within the Checker.destroy() method, and that is not called at all from the Eclipse plugin.
* Are there any hidden assumptions about the cachefile on the checkstyle core side? Is it okay to have the same checkfile for many checkers? The plugin creates one checker per eclipse project to be checked, and all share the identical cachefile right now.

### DIFF
--- a/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/builder/Auditor.java
+++ b/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/builder/Auditor.java
@@ -162,7 +162,21 @@ public class Auditor {
       // Cleanup listener and filter
       if (checker != null) {
         checker.removeListener(listener);
+        persistCacheFile(checker);
       }
+    }
+  }
+
+  private void persistCacheFile(Checker checker) throws CheckstylePluginException {
+    try {
+      // TODO use new API in checkstyle core once it might be available
+      var field = checker.getClass().getDeclaredField("cacheFile");
+      field.setAccessible(true);
+      var cacheFile = field.get(checker);
+      var persistMethod = cacheFile.getClass().getDeclaredMethod("persist");
+      persistMethod.invoke(cacheFile);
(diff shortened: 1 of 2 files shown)


rnveach:
One issue you will run into is https://github.com/checkstyle/eclipse-cs/issues/465 .

Eclipse-CS' current caching method is incomplete and causes issues. Adding a new cache on top of it will make figuring out caching issues harder to track.

I am all for more caching, but you may want to consider fixing the current caching before adding more to it.

Issue #437, opened 2022-12-31
upgrade to checkstyle 10.6.0


romani:
https://checkstyle.org/releasenotes.html#Release_10.6.0

PR #490, opened 2023-04-02
upgrade to 10.6.0

Bananeweizen:
This includes a new Ant script for updating the version. All changes were done by the Ant script, no manual changes at all.

@rnveach If I understand the release notes right, even though there is a breaking change, this does not affect the code itself. So this upgrade doesn't have any Java changes.

image


Bananeweizen:
fixes #437.

romani:
Please put link to issue to commit message

net.sf.eclipsecs-feature/pom.xml
```diff
@@ -3,7 +3,7 @@
     <parent>
         <artifactId>net.sf.eclipsecs.parent</artifactId>
         <groupId>net.sf.eclipsecs</groupId>
-        <version>10.5.0-SNAPSHOT</version>
+        <version>10.6.0-SNAPSHOT</version>
```
Calixte, line 6: I believe this was handled by the release script: https://github.com/checkstyle/eclipse-cs/blob/master/release.sh#L35
You probably will want to update it.

Calixte (approved):

net.sf.eclipsecs.checkstyle/META-INF/MANIFEST.MF
```diff
@@ -7,10 +7,10 @@ Bundle-Vendor: Eclipse Checkstyle Project
 Bundle-ActivationPolicy: lazy
 Bundle-RequiredExecutionEnvironment: JavaSE-11
 Export-Package: .,
- com.google.common.base;version="10.5.0",
- com.google.common.cache;version="10.5.0",
- com.google.common.collect;version="10.5.0",
- com.google.common.io;version="10.5.0",
+ com.google.common.base;version="10.6.0",
```
Calixte, line 10: The only difference with the PRs I usually make are those version numbers in Export-Package which I leave untouched and are bumped later by the release script.
I do not know what they refer to, though.

Calixte (approved):

romani:
Do we have a way to run update script from terminal ? By command

net.sf.eclipsecs.checkstyle/META-INF/MANIFEST.MF
Bananeweizen: Those are re-exports of the packages already exported by the checkstyle core library. This makes the libraries easily available in the eclipse-cs code, but there are multiple problems with this approach to my mind:
* the version numbers are plain wrong. Those are Google Guava packages, and guava has very different version numbers for its packages. That can easily lead to problems when other plugins also import the same packages and expect correct version numbers.
* the plugin should not depend on the dependencies of the core library in _its own code_. That way every upgrade of the core library has the potential to break code in the eclipse plugin for no good reason (e.g. if a method in Google Guava gets removed).

I'll create another change to get rid of these exported packages here, and will rebase this change afterwards.

Bananeweizen:
> Do we have a way to run update script from terminal ? By command

Not really sure, I've never tried to invoke an Ant script from the command line. However, I don't see that as necessary. Eclipse developers can just use the context menu "Run As" on that script and it will work fine, independent of OS and other installations. Eclipse has Ant support built-in.

net.sf.eclipsecs.checkstyle/META-INF/MANIFEST.MF
Bananeweizen: @Calixte The exports and the changed versions here are gone now. Thanks for the extra attention on the review and reporting this. Sometimes I'm quite surprised by what unusual things we find in the code here.

Calixte: It makes much more sense now, thank you.

Calixte:
I'll let you merge this one. Are you going to perform the release as well?

Bananeweizen:
> I'll let you merge this one. Are you going to perform the release as well?

@Calixte I would be more happy if you do it. I had a look at the release.sh script after you mentioned some version numbers being increased there. I'm on Windows only, and don't normally use any bash (also not the one coming with Git). That's why I'm not sure if I can get it running without making errors. In fact, I've been thinking whether that script should also be converted to something that can easily be executed cross-platform (e.g. adding just another Maven profile or Ant script doing all of that).

PR #497, opened 2023-04-07
Infra: make documentation timestamp reproducible

Bananeweizen:
Instead of adding the current time into the documentation sitemap, use the last change in any file of the docs project. Also use only the date part of the timestamp, not the full second resolution timestamp (using only the date is a valid W3C date format, which is used in sitemaps). That trick avoids vicious circle between committing documentation changes, updating the sitemap, amending to the last commit, updating the sitemap again etc.

That way rebuilding with maven from an existing git commit will no longer lead to unwanted changes in general.

@Calixte @rnveach When this has been merged and you see a locally modified sitemap.xml appearing in the future, then you should commit it, and no longer ignore/replace it, because then there have been changes in the documentation.

### DIFF
--- a/.ci/validation.sh
+++ b/.ci/validation.sh
@@ -4,20 +4,13 @@ set -e
 case $1 in
 
 git-diff)
-  printf "Changes to only docs/sitemap.xml:\n"
-  git diff docs/sitemap.xml
-
-  # file below is always modified on a run and updated with the current date
-  # such a change is ignored and must be verified manually instead
-  git checkout HEAD -- docs/sitemap.xml
-
   if [ "$(git status | grep 'Changes not staged\|Untracked files')" ]; then
     printf "Please clean up.\nGit status output:\n"
     printf "Top 300 lines of diff:\n"
     git status
     git diff | head -n 300
     false
-  fi 
+  fi
   ;;
 
 *)
--- a/net.sf.eclipsecs.doc/pom.xml
+++ b/net.sf.eclipsecs.doc/pom.xml
@@ -9,11 +9,31 @@
     <artifactId>net.sf.eclipsecs.doc</artifactId>
     <packaging>eclipse-plugin</packaging>
     <name>eclipse-cs Documentation Plugin</name>
-    <properties>
-        <timestamp>${maven.build.timestamp}</timestamp>
-    </properties>
     <build>
         <plugins>
+            <plugin>
+                <artifactId>maven-antrun-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>calculate-website-timestamp</id>
+                        <phase>process-resources</phase>
+                        <goals>
+                            <goal>run</goal>
+                        </goals>
+                        <configuration>
+                            <target>
+                                <exec outputproperty="doctimestamp" executable="git">
+                                    <arg value="log"/>
+                                    <arg value="-1"/>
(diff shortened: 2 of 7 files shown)


rnveach (changes requested):
> @Calixte @rnveach When this has been merged and you see a locally modified sitemap.xml appearing in the future, then you should commit it, and no longer ignore/replace it

Then it sounds like the following lines in the CI should be removed.
https://github.com/checkstyle/eclipse-cs/blob/master/.ci/validation.sh#L7-L12

This checks for uncommitted changes after validation run. It will fail if something is modified that should be committed and was not committed.
Just for note, this also means if something shouldn't be committed, like a temp file, then the CI command that gets executed needs to clean up after itself.

Bananeweizen:
> Then it sounds like the following lines in the CI should be removed.

Good catch. Removed that, hopefully the CI run is still happy with that script.

> Just for note, this also means if something shouldn't be committed, like a temp file, then the CI command that gets executed needs to clean up after itself.

I'm not aware of any temporary files that get created, except for those that are already ignored here: https://github.com/checkstyle/eclipse-cs/blob/master/.gitignore#L10. Do you know of any more?

rnveach:
> Do you know of any more?

Not for Eclipse. Main CS repo uses a lot more custom commands that also produce temporary files. That is where this check really shows these type of things. It was just added to Eclipse for safety.

net.sf.eclipsecs.doc/pom.xml
```diff
@@ -9,11 +9,31 @@
...
+                        <phase>process-resources</phase>
+                        <goals>
+                            <goal>run</goal>
+                        </goals>
+                        <configuration>
+                            <target>
+                                <exec outputproperty="doctimestamp" executable="git">
+                                    <arg value="log"/>
+                                    <arg value="-1"/>
+                                    <arg value="--pretty=&quot;format:%cs&quot;"/>
```
rnveach, line 28: It looks like `format` is being printed into the file on generation with quotes.

https://github.com/checkstyle/eclipse-cs/actions/runs/4645708368/jobs/8221697711?pr=497#step:6:26
-    <lastmod>2023-04-08</lastmod>
+    <lastmod>"format:2023-04-08"</lastmod>

.ci/validation.sh
```diff
@@ -4,20 +4,15 @@ set -e
 case $1 in
 
 git-diff)
-  printf "Changes to only docs/sitemap.xml:\n"
-  git diff docs/sitemap.xml
-
-  # file below is always modified on a run and updated with the current date
-  # such a change is ignored and must be verified manually instead
-  git checkout HEAD -- docs/sitemap.xml
+  git checkout HEAD
```
rnveach, line 7: I don't think the checkout is needed either. The original checkout was just to reset `sitemap.xml` when it got modified.

Bananeweizen: Now I get it. I had read the command as checkout everything except sitemap, which was completely wrong.

net.sf.eclipsecs.doc/pom.xml
Bananeweizen: Thanks for finding that. I now see that I can do the same without the additional quotes, as long as there are no spaces in the format.

rnveach (approved):

PR #498, opened 2023-04-07
Doc: checkstyle uses 3 part versions for all releases now

Bananeweizen:


Bananeweizen:
merged into other PR for simplicity

PR #502, opened 2023-04-08
infra: remove redundant docs maven build

Calixte:
I think the `mvn deploy -P release-composite` command will already generate the docs, so it's not useful to run maven again after.


romani:
It was added for bintray deploy https://github.com/checkstyle/eclipse-cs/commit/e2165d5bd5b5b0c9321deca35ec7eb9dc850712d

romani (approved):

PR #503, opened 2023-04-08
Infra: Filter derived files from own checks

Bananeweizen:
I just had some derived files (in the Eclipse bin directory) showing checkstyle issues. That's not useful, therefore enable the derived files filter for all projects.


rnveach:
This may not help for the maven run if the files are in it's path. Fixing it may depend on the exact file path.

Bananeweizen:
> This may not help for the maven run if the files are in it's path. Fixing it may depend on the exact file path.

This is not for the maven run, but really for checking files in the IDE used for development of eclipse-cs. I use the autoderiv plugin (see the .derived files), and that automatically marks several folders as derived (like bin or target), which then makes the filter exclude all contained files.

PR #516, opened 2023-04-14
Infra: more strict/reproducible Maven builds

Bananeweizen:
* add more Maven enforcer rules
* pin versions of all required default Maven plugins instead of relying on Maven super POM
* add maven-versions-plugin to plugin management, because that is used to verify and upgrade all these versions
* exclude more maven plugins from m2e execution, they are not needed for the eclipse build
* have all m2e lifecycle mappings in the parent pom, not in child modules

This PR also showcases the previous documentation timestamp changes: The first maven build with all the above changes committed produced new timestamps (because of docs/pom.xml being modified). The sitemap was then amended and will not change any further when building this commit again.

### DIFF
--- a/docs/sitemap.xml
+++ b/docs/sitemap.xml
@@ -15,77 +15,77 @@
   -->
   <url>
     <loc>https://checkstyle.org/eclipse-cs/</loc>
-    <lastmod>2023-04-15</lastmod>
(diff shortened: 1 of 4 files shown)


Bananeweizen:
(last commit was only rebase)

PR #517, opened 2023-04-14
Infra: Use Maven wrapper

Bananeweizen:
The Maven wrapper allows running Maven without it being installed, and to specify the version being used. That way all developers and CI systems can use the same version easily.

Travis supports mvnw:
https://docs.travis-ci.com/user/languages/java/#maven-dependency-management

As a developer, use "./mvnw goals" instead of "mvn goals" in the future, please.


Bananeweizen:
Even Travis is using Maven 3.9.1 (the latest release) with this approach. This can be seen by the new warnings like https://github.com/checkstyle/eclipse-cs/actions/runs/4698136013/jobs/8330060078?pr=517#step:5:149 which are only emitted in 3.9.1.

Bananeweizen:
(last commit was rebase only)

Bananeweizen:
@Calixte thanks a lot for review and merging. You can expect a slowdown now, vacation is almost over. ;)

Issue #518, opened 2023-04-14
outdated license files/headers


Bananeweizen:
At least all the LGPL license files and headers show an outdated address of the FSF. The current one is "51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA", as given in all their current license texts.

PR #519, opened 2023-04-14
Issue #518: update LGPL license header

Bananeweizen:
* address of FSF has changed
* replace manipulated license files (containing mixed whitespace) with a fresh copy from the FSF website.
* remove license file of removed component

This does not change any license version or any license text (except the address).

Also add a new checkstyle check to avoid the old address ever comes back.


config/checkstyle_checks.xml
```diff
@@ -159,6 +159,12 @@
                  value="All files in the ''src/it/java'' folder
                       should be named ''*Test.java'' or ''Abstract*.java''."/>
     </module>
+    <module name="RegexpSingleline">
+        <property name="id" value="oldFsfAddress"/>
+        <property name="format" value="\Q59 Temple Place\E"/>
+        <property name="message"
+                  value="The address of the FSF in licenses shall be ''51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA''."/>
```
romani, line 166: why we need this Check if we have update in `config/java_regexp.header`?
 any mismatch with header will be violation.

romani (changes requested):
items:

config/checkstyle_checks.xml
Bananeweizen: The same address exists in many more files than just Java files (e.g. properties, html, xml, ...). Since I'm not sure where people copy such license texts from (other than the FSF website with the correct address) I'd like to avoid someone creating another outdated license.html in the future.

romani (approved):
Ok to merge

romani (approved):
Ok to merge

config/checkstyle_checks.xml
Bananeweizen: It's not that people added the license as license header to these other files. It's just that other files contain pieces of the license text because licensing information is mandatory metadata e.g. for eclipse features or for the so called "branding plugin".

PR #532, opened 2023-05-25
infra: increase minimum required Eclipse version to 2021-06

### LINKED ISSUE #461: Development environnement issues with Apple M1
I'm unable to work on the plugin with my M1 mac.
I get the following error in CheckstyleMarkerImageProvider.java
> The type org.eclipse.swt.graphics.Image cannot be resolved. It is indirectly referenced from required type

I sometimes see the error when running `mvn install` as well but haven't been able to reproduce consistently. 

Architecture is arm64 or aarch64. 
I played with the tycho target platform configuration but haven't been able to get it to work yet.

Calixte:
Fixes #461

### DIFF
--- a/docs/partials/index.html
+++ b/docs/partials/index.html
@@ -8,7 +8,7 @@
                     <p style="margin-bottom: 5px;">
                         <a class="btn btn-primary btn-outline-inverse btn-lg" target="_blank"
                             href="http://marketplace.eclipse.org/marketplace-client-intro?mpc_install=150"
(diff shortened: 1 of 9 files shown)


Bananeweizen (changes requested):
The target change as such is okay. However, also the feature.xml should require the corresponding eclipse.core.runtime 3.22, otherwise people can still install the plugin in older unsupported versions.
And the documentation in install.html, index.html needs updating.

Calixte:
>  And the documentation in install.html, index.html needs updating.

Should the build script be changed to completely ignore the `docs` directory?
If I commit the changes to `docs`, then the website will be ahead of the actual release.

PR #474, opened 2023-02-05
config: adds git diff to github workflow

rnveach:
Adding git diff to ensure there are no differences after running install.

Expecting CI to fail currently without fix.

PR #540, opened 2023-08-15
Infra: ignore changes to docs directory

Calixte:
See https://github.com/checkstyle/eclipse-cs/pull/532#issuecomment-1615281172

### DIFF
--- a/.ci/validation.sh
+++ b/.ci/validation.sh
@@ -7,9 +7,9 @@ git-diff)
   printf "Changes to only docs/sitemap.xml:\n"
   git diff docs/sitemap.xml
 
-  # file below is always modified on a run and updated with the current date
+  # folder below is always modified on a run and updated with the current date
   # such a change is ignored and must be verified manually instead
-  git checkout HEAD -- docs/sitemap.xml
+  git checkout HEAD -- docs
 
   if [ "$(git status | grep 'Changes not staged\|Untracked files')" ]; then
     printf "Please clean up.\nGit status output:\n"


romani:
May be you need to just ignore spaces in diff?
So you will be notified only when content is changed.

Calixte:
I don't really understand why we have this validation script.
It forces us to commit the documentation changes to the `docs` directory ahead of the release instead of during the release. Sometimes we want to, sometimes not, so I'm not sure why we enforce it.
But anyways, everything is blocked by the Travis CI JDK 17 build that fails without any logs. Do we still need Travis now that we have GitHub CI?

romani:
I do not remember if Travis do something special, if we migrated completely off it, we can disable it.

For git change of this PR, I am to merge if you think there is no value in such git diff.

You can merge PR, even Travis is not happy.

romani:
I restarted Travis

rnveach:
Travis is just trouble and should be removed from all repos.

> if Travis do something special

Travis does JDK 17, but since it isn't working there is no point in keeping it. Just create an issue to restore it after it is removed.

> I don't really understand why we have this validation script.

@Calixte 

See https://github.com/checkstyle/eclipse-cs/pull/474#issue-1571560662

We use `git-diff` in main repo to ensure everything being modified by the script is either committed or removed to show nothing by the user was forgotten.

The little override being fixed here was there initially to basically ignore certain things changing since originally it was just timestamps. It is fine to ignore the whole folder if that is what is expected, the issue will only be at release time to remember to manually commit the folder with the release changes.

Calixte:
> the issue will only be at release time to remember to manually commit the folder with the release changes

The release script will commit any changes: https://github.com/checkstyle/eclipse-cs/blob/master/release.sh#L72

I'm going to make another PR to remove Travis.

romani:
If we ok to remove Travis, please merge this PR now.
State of Travis becomes irrelevant.

Bananeweizen:
Travis has been removed meanwhile, and all JDKs are tested at github, therefore merging.

PR #543, opened 2023-08-22
infra: remove travis ci

Calixte:
Is there anything else we need to do to remove Travis?


romani (approved):
Should be enough.
After merge, If not,  pin me I will do something on Travis web site to detach it from this repo

Bananeweizen:
the webhooks in the project were not removed. @rnveach did that yesterday.

PR #556, opened 2023-09-10
Add project configuration menu contribution

Bananeweizen:
That's the place where all project conversion and configuration contributions of different plugins should appear.
While that's basically a duplication of the sub menus in the Checkstyle command, it's the place where Eclipse developers might look first when trying to enable/disable checkstyle for a project.
The menu items are only visible if all selected projects do/don't have the checkstyle nature and are currently open.

The other menu items in this menu depend on what you have installed, e.g. you might see Xtext, Gradle, Groovy, Maven and other items appear there.
grafik


romani:
I highly recommend to keep "link" to PR or issue where more details (like image, ....) or conversation can happen and being useful to understand reasoning on code changes.
When we are lazy to create issue in main project, we create PR and change commit to reference PR number like "Pull #556: ......".

PR #566, opened 2023-09-17
Register quickfixes via extension point

### LINKED ISSUE #315: NeedBracesCheck: Quickfix is broken since 8.36.1
Since 8.36.1, you can no longer do a quick fix on `NeedBracesCheck` violations in the problem view. Quick fix is completely grayed out and does nothing when clicking.

Confirmed by installing 8.35 , verifying quick fix works, resetting eclipse (from pre-Checkstyle install save), installing 8.36.1, verifying quick fix does not work. Both versions showed the same checkstyle error after doing a clean build.

Input file:
````
package test;

/** Main. */
public class Main {
  /** main. */
  public static void main(String[] args) {
    // if (true)
    // System.out.println("test");
    if (true)
      System.out.println("test");
  }
}
````

Configuration: Google checks.

### LINKED ISSUE #561: performance issue when expanding checkstyle markers
Start an IDE which already has some checkstyle problem markers. Ensure to NOT trigger any scan, so the checkstyle modules metadata is not yet loaded. Expand the problem marker group containing the checkstyle markers (e.g. the "errors" group, or the "checkstyle issues" group, depending on how you have grouped them in the problems view).
This leads to blocking the IDE for as long as it takes to load all the metadata. This is because the problem marker images get a decoration which indicates if a marker has a quickfix. The information about quickfixes is attached to the metadata. This needs to be changed. Possible changes:
* have the quickfix itself contain the checkstyle module id that it can fix. That way loading of metadata is not necessary just for browsing the problem markers.
* have an OSGI declarative service triggered on "application started" lifecycle event (that means when Eclipse is up and running) and load the metadata concurrently at that time. This would mean loading the metadata as early as possible, therefore avoiding the delay later.

image

Bananeweizen:
* Remove old quickfixprovider extension point. It's entirely useless to register an OSGi plugin as extender of another plugin. OSGi will deal with that kind of class loading anyway.
* Instead have a new quickfix extension point, which registers a class implementing the quickfix related interface and the module that can be fixed this way.
* Completely avoids loading Checkstyle metadata for looking up quickfixes of existing problem markers.
* Fix ExplicitInitializationQuickfix throwing exceptions during the check whether it's applicable.

Fixes #315.
Fixes #561.

See the icons with the light bulb overlay here for violations with quickfixes available:
grafik


rnveach:
@Bananeweizen Just a question, does eclipse-cs have support for 3rd party (like sevntu) quick fixes where the code is in the other library? The way you add quickfixes here make me think that is a no, but I wanted to ask.

Bananeweizen:
Of course that works. As long as the quickfix is in a plugin, it will be found that way. _Us_ registering _our_ quickfixes is the same as someone else registering theirs. We only query the registered quickfixes, we don't have to deal with anything beyond that (class loading etc.). That's _the_ reason for having the extension point mechanism in the first place, to NOT deal with classloading on your own.

PR #569, opened 2023-09-19
Online help for problem markers

### LINKED ISSUE #568: Provide marker help
It's possible to register context sensitive help for each problem marker in the Eclipse help system. We can even calculate such online help entries dynamically and redirect them to the actual check description of the original project.

Bananeweizen:
* Register a problem marker help provider.
* Calculate a context help topic for each marker and dynamically forward it to the original online help of the related check.

For the end user, pressing F1 on a marker has the result of opening the integrated Eclipse help and showing the checkstyle problem description online.

Fixes #568.

See the modifier order problem marker selected and its help shown:
grafik

### DIFF
--- a/net.sf.eclipsecs.ui/src/net/sf/eclipsecs/ui/quickfixes/MarkerHelpContextProvider.java
+++ b/net.sf.eclipsecs.ui/src/net/sf/eclipsecs/ui/quickfixes/MarkerHelpContextProvider.java
@@ -1,0 +1,161 @@
+//============================================================================
+//
+// Copyright (C) 2003-2023 the original author or authors.
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+//
+//============================================================================
+
+package net.sf.eclipsecs.ui.quickfixes;
+
+import org.apache.commons.lang3.StringUtils;
+import org.eclipse.core.resources.IMarker;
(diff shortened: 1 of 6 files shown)


net.sf.eclipsecs.ui/src/net/sf/eclipsecs/ui/quickfixes/MarkerHelpContextProvider.java
```diff
@@ -0,0 +1,161 @@
...
+    if (metadata == null) {
+      return null;
+    }
+    var group = metadata.getGroup().getGroupId().toLowerCase();
+    // some web pages are different to the packages in Checkstyle
+    if ("indentation".equals(group)) {
+      group = "misc";
+    }
+    var file = moduleName.toLowerCase();
+    return "https://checkstyle.org/checks/" + group + "/" + file + ".html#" + moduleName;
```
rnveach, line 106: I am not sure all pages have the `#` and would ensure that is the case.

I know in the beginning there was discussions on this and I was against it since we only have 1 module per page now, so the `#` seemed excessively unnecessary and was really only skipping our page header. If we don't like our header that much, we might as well remove it.

rnveach: I am going to resolve this. I believe the actual issue was the links on our group pages. All the pages have the `#` because of how we define the page and the `#` creation.

Bananeweizen: Yeah, I think I know your confusion. I had similar when looking for the scheme. If you navigate from the TOC of the checkstyle page, the very first link of each group is different than the remaining ones. However, even for the first one of each group, the same link _exists_, so it can be used here.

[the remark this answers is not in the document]
Bananeweizen: AFAIK we will not be asked for a context in line 80 if we returned false from line 70. E.g. first comes the "hasContext", and only then the "getContext" call (if hasContext was true). But I could add a second identical check in getContext to be on the safe side. We will not get into this method (with your comment) if getContext() doesn't return a context.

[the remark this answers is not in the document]
rnveach: I will take your word at this, as I don't know EclipseCS enough on how context connects to if the check if from checkstyle or a 3rd party. I was just raising the concern that the link create will be invalid if you run it for a 3rd party check as it's documentation will not be on Checkstyle's website.

PR #581, opened 2023-09-23
Ignore checkstyle dependency update

Bananeweizen:
* ignore the dependency to the main project in dependabot, because we have to do these changes manually
* remove quotes, gives better syntax highlighting in VS code
* also check the github actions

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,7 +3,13 @@
 version: 2
 updates:
   - package-ecosystem: maven
-    directory: "/"
+    directory: /
     schedule:
-      interval: "weekly"
-    open-pull-requests-limit: 10
+      interval: weekly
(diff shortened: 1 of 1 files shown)


Bananeweizen:
@rnveach I created this change because our POM says that we need to have a compatible version for running the sevntu checks and the patch filter. However, the dependabot upgrade PR of this dependency was okay. Now I'm a bit puzzled, whether or not we can upgrade that 10.4.0 version. Do you know better?

rnveach:
I don't recommend breaking version compatibility with the others that depend on it. While yes, its possible everything can be totally fine if it is manually tested first, they could actually mash against each other and result in exception, or worse false negatives. I thought when I first pointed this out when were enabling it, there was some issue that went unnoticed.

PR #582, opened 2023-09-23
infra: downgrade JUnit

Bananeweizen:
Upgrading Maven Surefire to a recent version fails with exceptions in the process launching the test process. That's because we have 2 versions of JUnit: 5.10 from Maven and 5.7.1 from the Eclipse target platform. Surefire seems to be confused by that and tries to use the junit-platform-launcher 5.7 from the target platform to launch the 5.10 JUnit during the Maven run.

To avoid this, downgrade the Maven JUnit version to the latest 5.7.x, such that both Maven and Eclipse use the same JUnit version.

Also remove some superfluous configuration.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,7 +22,9 @@
         <build-helper-maven-plugin-version>3.4.0</build-helper-maven-plugin-version>
         <download-maven-plugin-version>1.7.1</download-maven-plugin-version>
         <exec-maven-plugin-version>3.1.0</exec-maven-plugin-version>
-        <junit-jupiter-version>5.10.0</junit-jupiter-version>
+
+        <!-- Maven JUnit version must fit the JUnit version used in the target platform, otherwise Surefire may produce exceptions -->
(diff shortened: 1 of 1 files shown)


Calixte:
> 5.7.1 from the Eclipse target platform

Out of curiosity, how do you check the junit version of the eclipse target platform?

Bananeweizen:
@Calixte If you already have a target platform loaded, then ctrl-3, target platform state gives you a view with the bundles contained in the target, and you can open them as if they were workspace bundles.
If you are still in the creation of the target platform, the TPD editor has both code completion for adding the version after the bundle/feature name, and it also has a hover which displays the version when hovering of a bundle name.
Also if you go the project explorer, into one of your plugins and then expand either the node "Plugin dependencies" or "JUnit 5", you would see the bundles coming from the manifest or from the JUnit container of Eclipse with their specific versions.
And when launching a plugin test launch config (plugin test, not junit test), then the "plugins" tab also shows the versions of the bundles taking part in the launch.
There are probably some more ways... ;)

PR #587, opened 2023-09-25
Upgrade to 10.10.0

### LINKED ISSUE #524: upgrade to checkstyle 10.10.0
https://checkstyle.org/releasenotes.html#Release_10.10.0

Bananeweizen:
The new filter works, but every adding/removing of a filter comment only hides/re-adds a violation after purging the Checkstyle cache, not on save like for any other file content change. I'm not sure if that is a problem in the main checkstyle part or in Eclipse-CS. I'd still suggest to commit and release the upgrade and to investigate why cache purging is needed only later.

Fixes #524.

grafik

grafik


romani:
Please link issue to cache, to make it clear we see issue and it is registered and will be handled separately.

Bananeweizen:
Only afterwards I noticed that @Calixte had already described that same problem in the initial issue: https://github.com/checkstyle/eclipse-cs/issues/524#issuecomment-1696156087

PR #592, opened 2023-09-29
Add proper permission for publish-unit-test-result-action

Calixte:


Calixte:
> @Calixte Just curious, how did you know these permissions were needed and which were needed?

It's in the README of that action: https://github.com/EnricoMi/publish-unit-test-result-action#permissions

Bananeweizen:
@Calixte thanks for looking into that. While I have done my share of Maven and Eclipse, I'm fairly new to GitHub actions. I was also wondering if there is a better way to test such changes than just creating the PR. To my understanding that is sometimes a kind of chicken-and-egg situation when creating a change on an action that should verify a PR, but that PR itself is of course not yet using that changed code (or is it)?

rnveach:
I don't think the PR is using it. That would be a security issue if someone malicious sent a PR.

BTW, @Bananeweizen @Calixte please read https://thehackernews.com/2023/09/github-repositories-hit-by-password.html?m=1 .

PR #598, opened 2023-10-17
infra: harden GitHub workflows

Bananeweizen:
See https://github.com/step-security/secure-repo#1-automatically-set-minimum-github_token-permissions for details. Dependabot will be able to upgrade the pinned dependencies and will use the same format (including version comments).

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -15,12 +15,15 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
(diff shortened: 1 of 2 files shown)


rnveach:
@romani @nrmancuso ping for main repo's GH workflows

rnveach:
@Bananeweizen Just a question, do you know if this has any impact on dependabot updates for GH actions?

I ask because you specify the commits, but when we enabled dependabot for these in our repo, it updated version numbers. Not sure if some other configuration is needed, or if it will work out of the box here.

rnveach:
Nevermind, missed your comment completely.

Bananeweizen:
recreated as #611 instead

PR #613, opened 2023-12-20
Fix build order

Bananeweizen:
The unpacking previously happened in the prepage-package lifecycle phase of maven. That's far too late, since other compilation steps require those sources. I had multiple failing local builds because of this.

Until now this only worked correctly because either
* the eclipse builder would compile the depending sources again when the maven build unpacked everything
* the maven build was not fast enough in parallel processing
* the sources were still available locally from the last download, when not using clean in between

Also ensure the order of download and unpack is correct by converting the download to an Ant task and moving it into the same execution. As a side effect, this avoids the download progress messages of the previous download-maven-plugin.

### DIFF
--- a/net.sf.eclipsecs.checkstyle/pom.xml
+++ b/net.sf.eclipsecs.checkstyle/pom.xml
@@ -9,40 +9,28 @@
     <artifactId>net.sf.eclipsecs.checkstyle</artifactId>
     <packaging>eclipse-plugin</packaging>
     <name>Checkstyle Core Library Plugin</name>
-    
+
     <properties>
         <checkstyle.version>10.10.0</checkstyle.version>
     </properties>
     <build>
         <plugins>
             <plugin>
-                <groupId>com.googlecode.maven-download-plugin</groupId>
-                <artifactId>download-maven-plugin</artifactId>
-                <executions>
-                    <execution>
-                        <id>download-checkstyle-src</id>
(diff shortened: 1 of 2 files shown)


net.sf.eclipsecs.checkstyle/pom.xml
```diff
@@ -9,40 +9,28 @@
...
                 <artifactId>maven-antrun-plugin</artifactId>
                 <executions>
                     <execution>
-                        <phase>prepare-package</phase>
+                        <id>repackage-checkstyle-source</id>
+                        <phase>generate-sources</phase>
                         <configuration>
                             <target>
+                                <get src="https://github.com/checkstyle/checkstyle/archive/checkstyle-${checkstyle.version}.zip"
+                                     dest="${basedir}/checkstyle-${checkstyle.version}.zip"
```
rnveach, line 28: Shouldn't this really go to the target folder instead? Doesn't base dir put it into the path of what gets cataloged for git, so this would show up as an unchanged file?

Bananeweizen: Completely right, and I was thinking the same. However, it has been like that since eternity. I wasn't sure if I would break some of the scripts of calixte when moving this around, therefore I would like to do that as additional cleanup later (also looking at other places where files might be created outside the maven target directory).

rnveach (approved):

PR #614, opened 2023-12-20
Upgrade settings and enforce common import order

Bananeweizen:
Upgrade the JDT settings to the current 2023-12 values (there are some new automated cleanups). Enable a custom import order in every Java project, because there are different orders in place right now (sometimes net.sf comes first, sometimes not).

Organize imports in all Java source files to enforce the new order.

The only manual changes are in the second commit.


Bananeweizen:
@rnveach any idea why the patch filter itself (not its tests!) fails during the PR run?

rnveach:
CI was recently updated in https://github.com/checkstyle/eclipse-cs/pull/611 . f15d6d85808eadddfa0d9a6ae8994a950528ff28 was merged after that, but I couldn't tell if it was using an older version before the CI update. This doesn't really narrow anything down, but this shows the CI environment recently changed.

Bananeweizen:
@rnveach thanks for all the details. A line ending issue or similar sounds reasonable. On my local system running mvn verify is completely fine. The CI change that you point to is a non-functional change. The git actions in there are exactly the same as before, they are just addressed by their git commit hashes and not their versions (but their version tags resolve to those revisions).

Given that the patch mechanism is basically no longer relevant after a change is merged, can we eventually just ignore the failure? It would probably appear again on the master build, but should not appear anymore once the next PR is merged, right?

rnveach:
> Given that the patch mechanism is basically no longer relevant after a change is merged, can we eventually just ignore the failure? It would probably appear again on the master build, but should not appear anymore once the next PR is merged, right?

If master is set to run the same maven commands, it should be running the same git commands which compares HEAD to previous HEAD. Unless CI does some weird things to the git repository which travis and some others did, then I would assume it will fail again on merge.

I agree it shouldn't appear on next CI run, unless you run into the exact same issue which we haven't fully identified what it is.

I at least recommend that we capture the patch file from this PR failure, and try some things with it like using it on our locals to reproduce, and comparing our local patch file with it and do a text/hex compare to see what the actual difference is.

The best way to capture the file is to have GitHub's CI log the file away for storage, which will capture it in it's binary state and not worry about losing anything to file conversions.
Example: https://github.com/checkstyle/checkstyle/blob/master/.github/workflows/pitest.yml#L91-L99

PR #615, opened 2023-12-20
infra: clean up build goals

### LINKED ISSUE #616: patch for sevntu is created for each module
The patch creation for the sevntu checks is executed once per module. That will be more visible once #615 is merged. That's not really necessary, or is @rnveach ? To my understanding the patch is created for the complete repository, independent of the modules. I guess that it would be okay if I change the build to only create that patch once, in an arbitrary module (which could be done via a maven profile with a file existence condition).

Bananeweizen:
* disable a tycho goal which we don't need
* execute the patch creation only once overall, not for each module

Fixes #616

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -138,27 +138,6 @@
                 </configuration>
             </plugin>
             <plugin>
-                <groupId>org.codehaus.mojo</groupId>
-                <artifactId>exec-maven-plugin</artifactId>
-                <executions>
-                    <execution>
-                      <phase>verify</phase>
-                      <goals>
-                        <goal>exec</goal>
-                      </goals>
-                    </execution>
-                </executions>
-                <configuration>
-                    <executable>git</executable>
-                    <arguments>
-                      <argument>diff</argument>
-                      <argument>HEAD~1</argument>
-                      <argument>HEAD</argument>
-                    </arguments>
-                    <outputFile>${maven.multiModuleProjectDirectory}/show.patch</outputFile>
-                </configuration>
-            </plugin>
-            <plugin>
                 <artifactId>maven-checkstyle-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -143,18 +143,19 @@
                 <artifactId>exec-maven-plugin</artifactId>
                 <executions>
                     <execution>
-                      <phase>verify</phase>
-                      <goals>
-                        <goal>exec</goal>
-                      </goals>
+                       <id>create-sevntu-patch</id>
```
rnveach, line 146: Like we said in the issue, it isn't really sevntu creating this, and this isn't used by sevntu but [patch-filters](https://github.com/checkstyle/patch-filters/blob/master/pom.xml#L13).

Since this is created for the whole repo, maybe it is better to call it something like `create-git-patch-file`?

pom.xml
```diff
@@ -143,18 +143,19 @@
...
                 <configuration>
                     <executable>git</executable>
                     <arguments>
-                      <argument>diff</argument>
-                      <argument>HEAD~1</argument>
-                      <argument>HEAD</argument>
+                       <argument>diff</argument>
+                       <argument>HEAD~1</argument>
+                       <argument>HEAD</argument>
                     </arguments>
```
rnveach, line 159: This is now +3 over `</arguments>`, but `<arguments>` is +4 over `<configuration>`. `<executions>` is +4 over `<execution>`. I would make this all consistent.

Bananeweizen: Not sure how that escaped my 👀 . Fixed. In fact, I now looked up the Maven coding conventions, and they want 2 spaces. I'll put automated formatted of the POM on my todo list. :)

rnveach (approved):

PR #631, opened 2024-01-21
Replace quickfix GIFs by PNGs

Bananeweizen:
Also add a more specific one for the NeedBraces quickfix.

There is no functional change, just better support for HiDPI displays.


rnveach:
@Bananeweizen Isn't quickfixes still broken at https://github.com/checkstyle/eclipse-cs/issues/315 .

If so, then I would question how you can be sure this doesn't break anything for something that is already broken.

Calixte:
> @Bananeweizen Isn't quickfixes still broken at #315 .
> 
> If so, then I would question how you can be sure this doesn't break anything for something that is already broken.

#315 is fixed by #566 so I was able to verify this works.

rnveach:
I forgot that there was a PR sitting for it to be merged.

PR #636, opened 2024-01-25
normalize feature.xml to current format

Bananeweizen:
The feature.xml serialization was changed in Eclipse 4.30: https://eclipse.dev/eclipse/news/4.30/pde.php#unnecessary-attributes-removal

This change just stores the feature.xml without any changes, to avoid confusion in the future by other editors when doing actual changes.


Bananeweizen:
@Calixte You would get the same changes in your IDE (assuming it is the 2023-12 release) if you do any change to the file from within the feature editor. So you could of course include this in your next version upgrade, if you want to (and delete this PR).

PR #659, opened 2024-03-09
Lift Java version requirement

Bananeweizen:
The minimum version of 17 is needed since a while and should have been changed with the Tycho 4 upgrade already. The previous limit of the upper version is no longer needed, all involved tools work fine with 21.

Noticed because my personal default Java is 21 meanwhile, and the build would stop there, for no good reason.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -392,7 +392,7 @@
                                     <banDuplicatePomDependencyVersions/>
                                     <requireExplicitDependencyScope/>
                                     <requireJavaVersion>
-                                        <version>[11,18)</version>
+                                        <version>[17,)</version>
                                     </requireJavaVersion>
                                     <requireMavenVersion>
                                         <version>${maven-version}</version>


rnveach:
If you want and is supported, a higher jdk can be added to show it is supported too.
https://github.com/checkstyle/eclipse-cs/blob/master/.github/workflows/maven.yml#L12

Bananeweizen:
While that's technically possible (see https://maven.apache.org/enforcer/enforcer-rules/versionRanges.html) it's very uncommon to have something that is not an open/closed interval, and some tools might stumble then. In fact, that's exactly the root cause for https://github.com/checkstyle/checkstyle/issues/14376. So I would really leave this as an open interval instead.

rnveach:
> I would really leave this as an open interval instead.

I was speaking of the CI, not the POM, to showcase the upper JDK is supported. Sorry if I wasn't clear.

romani:
Keep moving forward to jdk 17, do not wait for main library to update, we will do it very soon.

Bananeweizen:
> I was speaking of the CI, not the POM, to showcase the upper JDK is supported. Sorry if I wasn't clear.

Ah, got that wrong then. https://github.com/checkstyle/eclipse-cs/pull/660
Also if my comment on this PR isn't clear, the Java 17 requirement is only for the Maven JDK, not for the Eclipse plugin itself.

> Keep moving forward to jdk 17, do not wait for main library to update, we will do it very soon.

Great news. I'll create another change for that. If you are doing some braking changes anyway, would you also consider removing Guava completely? Checkstyle is a bad guy in that regard: https://twitter.com/aalmiray/status/1757710195201823125. The Eclipse plugin is almost free of Guava on my local branch, so the plugin would not be affected by the core project removing it.

rnveach:
> If you are doing some braking changes anyway, would you also consider removing Guava completely?

I doubt we will be doing breaking changes for the java upgrade.

When removing Guava only 2 files break. One seems very minor, but the other is `ModuleReflectionUtil`. [We use Guava for our classpath scan to identify and pull in 3rd party modules we find on the path](https://github.com/checkstyle/checkstyle/blob/b518572536891f3f3a327c42e54b9f3410d0cef7/src/main/java/com/puppycrawl/tools/checkstyle/utils/ModuleReflectionUtil.java#L59-L62). We would need to find something similar to continue this functionality if we were to remove Guava.

This is besides the fact that we would need an approved issue to do this removal.

romani:
As far as I remember, this is the only usage of guava in our project, we need to find alternative to remove guava.

rnveach:
Someone needs to create an issue, approve it and move it forward, otherwise nothing will happen. Like I said there are 2 files that need to be modified and an alternative is needed. We also need to remove reflections library as since move from Java 8, it hasn't been updated and basically does nothing. https://github.com/ronmamo/reflections/issues/465 https://github.com/ronmamo/reflections/issues/186

PR #667, opened 2024-03-17
Fix new violations

Bananeweizen:
These violations are shown with the most recent 10.14.1 by the eclipse plugin itself (because there was a fix in the "maximum distance between declaration and first usage" check), but are not part of the github verification due to an older version being used there.
@rnveach I guess we still can't do anything about upgrading that 10.4 version in the build? I don't remember exactly but I believe you had told me about either the filters or sevntu project being an issue there.


rnveach:
The one connected to patch-filters. The library must be upgrade, and sevntu must be upgraded first.

If this is an issue, then I would recommend patch-filters disconnects itself from sevntu and ask they keep regular updates. It is that or merge it into the main repo and keep it updated there.

The only other options is to do all these other library upgrades ourselves or to remove patch-filter altogether as it isn't being properly maintained.

PR #671, opened 2024-03-20
resolve OverridableMethodInConstructor violations

Calixte:


net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/config/ConfigProperty.java
```diff
@@ -46,10 +46,9 @@ public class ConfigProperty implements Comparable<ConfigProperty>, Cloneable {
    *          the property meta data
    */
   public ConfigProperty(ConfigPropertyMetadata metaData) {
-
     this(metaData.getName(), metaData.getOverrideDefault() != null ? metaData.getOverrideDefault()
             : metaData.getDefaultValue());
-    setMetaData(metaData);
+    this.metaData = metaData;
```
Bananeweizen, line 51: I'm not a friend of inlining setters. The moment someone adds an additional line in a setter, like a precondition, a null check etc., all the inlined duplications of it are wrong, because they miss that same change. Any reason why you are doing that?

Bananeweizen (commented):
While I do like the removal of the "m" prefixes, I don't like inlining setters. And I also don't understand were the violations are coming from. I don't have violations using the most recent release of our plugin. Did you use an even more recent version? Or did you add a check in the config and missed committing that config change?

Calixte:
> I also don't understand were the violations are coming from. I don't have violations using the most recent release of our plugin. Did you use an even more recent version? Or did you add a check in the config and missed committing that config change?

We have this check configured in our config: https://github.com/checkstyle/eclipse-cs/blob/master/config/checkstyle_sevntu_checks.xml#L101
You don't see violations because we use patch-filter to suppress the violations if they are in old code: https://github.com/checkstyle/eclipse-cs/blob/master/config/checkstyle_sevntu_checks.xml#L21
If we enforce a rule in new code, we should thrive to eventually fix old code and this PR is an attempt at making progress.

net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/config/ConfigProperty.java
Calixte: I think there's a difference between inlining setters and moving away from using setters in constructors, which is what I'm doing here.
Usually, in the body of a method, you should be able to assume you have a fully constructed object. Calling methods in constructors is therefore not ideal.
If I had validation logic I wanted applied in the constructor, and it was worth sharing it with the setter, I'd make a private method that performs this validation.
I'm not worried about someone "miss[ing] that same change", because it would be bad to assume that all constructors are all calling the setters: it's not generally the case.

The alternative to this is just to finalize the setters used in constructors, which I've done once in this PR, but I think not using setters in constructors at all is a better solution.

Bananeweizen:
Okay, I can accept that. But please be aware that for a lot of the classes that subclass Eclipse API (like wizards or dialogs) this will not be possible, because the super classes are sometimes written in a way that enforces the subclass constructor to also call the method that the super class constructor calls.

Regarding general cleanup: If we want to really get rid of violations in old code, then we should also apply everything that can be fixed automatically, instead of manually fixing one rule after the other. I know that I can fix more than 10.000 violations of the overall rules in 1 day by just using the Eclipse built-in code cleanups and formatter settings (when I tried that last time, it fixed 16.000 violations). However, since that will touch basically all files and a lot of lines, our own checks would then report all _the other remaining_ issues on those touched lines as "new code" violations. To my mind all those "new, but not really new" violations should simply be ignored for that one cleanup PR, as it will not be possible to fix them all manually in the same commit. I've already suggested that in https://github.com/checkstyle/eclipse-cs/pull/435, but could not convince others. If you see value in such an approach, then I could just do that again over the Easter holiday. I don't intend to ever fix those 10.000+ issues manually.

Bananeweizen (approved):

Calixte:
> I've already suggested that in https://github.com/checkstyle/eclipse-cs/pull/435, but could not convince others.

That PR was merged eventually, so I don't think we have so many violations left in the code? There's a few left and I was going to get to them with a mix of automated and manual efforts. I don't mind spending a bit of time doing them because it's a good excuse to get familiar with the code base.

PR #675, opened 2024-03-26
use tycho-eclipse-plugin instead of tycho-eclipserun-plugin

Calixte:
See https://github.com/checkstyle/eclipse-cs/commit/49972e4099b4899e402c57ce3bd651fb1517c9e0#r140269988

### DIFF
--- a/net.sf.eclipsecs-updatesite/pom.xml
+++ b/net.sf.eclipsecs-updatesite/pom.xml
@@ -131,7 +131,7 @@
                     </plugin>
                     <plugin>
                         <groupId>org.eclipse.tycho</groupId>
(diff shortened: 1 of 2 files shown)


Bananeweizen (approved):
Sorry, seems I missed that. I never really know whether I can test the publishing locally without causing complete chaos for the online update site etc.

Bananeweizen:
I'll merge this one, because I have further maven changes upcoming and would like to the see the warnings for the missing plugins be removed first. If this doesn't work as intended for the release, then feel free to revert the part of my commit which renamed the eclipserun plugin from tycho.extras.

PR #681, opened 2024-03-31
only scan relevant classpaths for metadata

### LINKED ISSUE #515: Huge performance gap due to reading unnecessary metadata
#242 introduced some eclipse-metadata.yml. To my understanding that is not needed at all. AFAIK only the sevntu checks have that yaml file in their jar (because it was introduced with that PR). And even there it's not needed, because the sevntu modules are perfectly found and displayed by the normal checkstyle-metadata. Deleting that yaml related code doesn't make any difference for me. (I've debugged the entire metadata loading with the sevntu eclipse plugin in my workspace, and I can configure and execute them as before).

At the same time, scanning all classes on the classpath for those yaml files takes an eternity and is the main responsible for the slow initialization of metadata on first access to any rule (remember #375). Below image shows this taking 27.8 seconds of CPU time on my performance laptop.
grafik

Am I missing anything about those yaml files or can we really just get rid of this?

Bananeweizen:
* don't scan the com.* package hierarchy
* calculate the common root packages of all known packages with checks and metadata instead
* scans only the ...puppycrawl... and net.fs...samplechecks packages in a default developer runtime now

This still finds all relevant metadata files, but takes 480 ms instead of 12 seconds on my laptop.

fixes #515

### DIFF
--- a/config/checkstyle_checks.xml
+++ b/config/checkstyle_checks.xml
@@ -396,6 +396,12 @@
             <message key="import.illegal"
                      value="Use org.assertj.core.api.Assertions instead."/>
         </module>
+        <module name="IllegalImport">
+            <property name="illegalPkgs"
+                      value="org.reflections"/>
+            <message key="import.illegal"
+                     value="Use io.github.classgraph.ClassGraph instead."/>
+        </module>
         <module name="ImportControl">
             <property name="id" value="ImportControlMain"/>
             <property name="file" value="${checkstyle.importcontrol.file}"/>
--- a/net.sf.eclipsecs.core/META-INF/MANIFEST.MF
+++ b/net.sf.eclipsecs.core/META-INF/MANIFEST.MF
@@ -9,7 +9,8 @@ Bundle-ClassPath: .,
 Bundle-Activator: net.sf.eclipsecs.core.CheckstylePlugin
 Bundle-ActivationPolicy: lazy
 Bundle-Vendor: %Bundle-Vendor
-Require-Bundle: net.sf.eclipsecs.checkstyle;visibility:=reexport
+Require-Bundle: net.sf.eclipsecs.checkstyle;visibility:=reexport,
+ io.github.classgraph.classgraph;bundle-version="4.8.168"
 Eclipse-BuddyPolicy: registered
 Export-Package: net.sf.eclipsecs.core,
  net.sf.eclipsecs.core.builder,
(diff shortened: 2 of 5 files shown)


rnveach:
You confirm it still finds and loads sevntu's YML?
https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/eclipsecs-sevntu-plugin/src/com/github/sevntu/eclipse-metadata.yml

rnveach:
Unrelated but `Reflections` should be removed. It doesn't work as expected on newer JREs and the repo is abandoned. I brought this up before and to main repo as well.
https://github.com/checkstyle/eclipse-cs/pull/659#issuecomment-1987274759

Bananeweizen:
@rnveach It finds that package when the sevntu plugin is running: 
grafik

However, _neither_ on master nor on my branch _any_ metadata file is found by scanning the classpath.

Regarding the reflections library: If you know a good alternative, then please raise an issue for replacing reflections.org. Don't be confused, I will probably _add_ it to the target platform in another PR soon, where I'm working on consuming the checkstyle.jar instead of the checkstyle-all.jar. Adding it is only to move the dependency to this project instead of taking it transitively from checkstyle.

rnveach:
> However, neither on master nor on my branch any metadata file is found by scanning the classpath.

I assume this is related to other issues/comments on the matter as sevntu has been broken for a while. I am just confused if you say it won't load the YML or if it won't load something else?

> If you know a good alternative

I believe Reflections right now is just doing the very basic load (common java code), so I would recommend doing that. No extra library needed. I do not remember the specifics as it has been a while since I originally looked into it.

Bananeweizen:
> I am just confused if you say it won't load the YML or if it won't load something else?

It doesn't _find_ any YML. So there is no loading.

rnveach:
Thanks. So this probably related to the reflections library. Since it doesn't work properly in Java 9+, it isn't finding anything.

Bananeweizen:
@rnveach I've changed the implementation, and now the classgraph library is used instead. It finds the sevntu checks, but unfortunately it still uses 7 seconds of CPU time. Nevertheless I consider this change fixing both the performance gap as well as the reflections issue.

Bananeweizen:
I'm going to merge this and the 2 PRs that are based on it. Feel free to still review and comment, so I can address comments separately, if necessary. However, I have some more classpath related changes (like trying to use the slim checkstyle.jar and to remove the guava dependencies) coming up, and all these changes are in conflict with each other because they change how the different plugin projects depend on each other.

PR #684, opened 2024-04-01
Add sevntu to target platform

Bananeweizen:
This adds the sevntu plugin to the target platform. That way launching a debug runtime from the IDE automatically includes sevntu. Have a checkstyle import check to avoid anyone referencing code from the sevntu plugin.

Be aware that this PR includes #681, because changes on the generated .target file are _always_ conflicting. Therefore 681 should be merged first, and then this PR.

### DIFF
--- a/config/checkstyle_checks.xml
+++ b/config/checkstyle_checks.xml
@@ -402,6 +402,12 @@
             <message key="import.illegal"
                      value="Use io.github.classgraph.ClassGraph instead."/>
         </module>
+        <module name="IllegalImport">
+            <property name="illegalPkgs"
+                      value="com.github.sevntu"/>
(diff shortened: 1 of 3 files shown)


rnveach:
> That way launching a debug runtime from the IDE automatically includes sevntu

Just confirming, this only includes sevntu for debugging, it will not include sevntu in the eclipse-cs release?

Bananeweizen:
> Just confirming, this only includes sevntu for debugging, it will not include sevntu in the eclipse-cs release?

Exactly. It would have to be included into the feature.xml to become part of eclipse-cs. And I added the checkstyle rule to avoid any accidental reference from eclipse-cs to sevntu code.

PR #685, opened 2024-04-01
Remove duplicate checkstyle.jar

### LINKED ISSUE #385: duplicate nested checkstyle jar
The jar containing the main checkstyle code exists two times in eclipse-cs. Once in eclipsecs.core/lib, second time in eclipsecs.checkstyle/. To my understanding the eclipsecs.checkstyle project is meant to be the container for that lib, therefore we should remove it from eclipsecs.core/lib. Or is there any good reason to have it 2 times, that I'm not aware of?

Bananeweizen:
* remove and gitignore the checkstyle.jar in eclipsecs.core
* have only the copy in eclipsecs.checkstyle remain
* export packages of transitive dependencies from the eclipsecs.checkstyle project. this is not a good practice, but represents exactly what happened before: consuming the third party dependencies from the checkstyle library

fixes #385

In a follow-up change the re-exported packages should be removed, and instead be replaced by taking the respective libraries into the target platform.

Please be aware this change is based on #684 and #681, therefore those should be merged first.

### DIFF
--- a/net.sf.eclipsecs.checkstyle/META-INF/MANIFEST.MF
+++ b/net.sf.eclipsecs.checkstyle/META-INF/MANIFEST.MF
@@ -7,6 +7,9 @@ Bundle-Vendor: %Bundle-Vendor
 Bundle-ActivationPolicy: lazy
 Bundle-RequiredExecutionEnvironment: JavaSE-17
 Export-Package: .,
+ com.google.common.base,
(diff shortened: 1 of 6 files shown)


Bananeweizen:
Merging, so I can continue with other PRs that modify the classpath and dependency chain. Feel free to still raise issues or comments.

PR #687, opened 2024-04-02
Revert "remove duplicate checkstyle.jar"

Bananeweizen:
This reverts commit d74a25f7f64ccaf650d11fa3431e2e2b4216f21f.

The removal of the second copy of the checkstyle.jar worked fine, no compiler issues, no failing tests, and all checks working as intended. However, the configuration dialog didn't display module configurations any more. There is a problem with the re-exported package for the metadata, specifically this line: https://github.com/checkstyle/eclipse-cs/blob/33ce44e8801d824387d4b90bad1e8c096a5ece4c/net.sf.eclipsecs.checkstyle/META-INF/MANIFEST.MF#L37

I have to investigate more deeply how to make the classpath work correctly without re-exporting that package. Until then, revert the commit, and also add a new check that will throw a runtime error, if no module metadata can be read.


Bananeweizen:
I'm going to merge this, so the broken functionality on master is repaired. Sorry for the noise.

PR #691, opened 2024-04-04
Add command for running checkstyle

### LINKED ISSUE #99: Add missing keybinding for "Check Code with CheckStyle" command
Currently, there is no way to tie this command to a shortcut key because the key binding for it isn't defined in plugin.xml. 

I've added the modifications to plugin.xml to bind the "Check Code With CheckStyle" command to a key that is configurable under Eclipse > Preferences > Keys and default it to M1+M2+K (ex: CTRL+ALT+K). This way, users can quickly CheckStyle on the active editor (or files selected in the Package Explorer) without having to use the context menus or rely on building. 

I'll commit the changes if this issue is approved.

Bananeweizen:
This adds a command for the already existing action to run checkstyle on the selected resources. That way users can invoke the command also via Ctrl-3 quick assist or via a key binding.

There is no default key binding, as the normal workflow for users is to rely on the Eclipse builder and to not invoke this command manually.

fixes #99

grafik

Be aware the shown keybinding is not available by default. I just added that for testing and to illustrate that a keybinding _can_ be given by the user.

### DIFF
--- a/net.sf.eclipsecs.ui/plugin.xml
+++ b/net.sf.eclipsecs.ui/plugin.xml
@@ -308,7 +308,13 @@
         point="org.eclipse.ui.commands">
         <category id="net.sf.eclipsecs.ui" name="%commands.category.name"/>
         <command
-            categoryId="org.eclipse.jdt.ui.category.source"
+            categoryId="net.sf.eclipsecs.ui"
+            description="%CheckSelectedFilesAction.tooltip"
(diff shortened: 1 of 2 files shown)


romani:
@Bananeweizen , please do not merge your own PRs , please wait for review from someone else.
PR is opportunity for others to understand details on how it works and ask questions.
We should avoid code that only single person understands.

If PR is hanging for long without review, pin me or @rnveach , we will merge.

PR #707, opened 2024-06-03
maven cleanup

Bananeweizen:
* unbind the default tycho-bnd lifecycle binding, we don't use any bnd instructions
* require same version for all tycho plugins
* exclude eclipsecs.checkstyle from source feature build


romani:
@Calixte , please review

romani:
@Calixte , please review

PR #708, opened 2024-06-03
upgrade external target platform dependencies

Bananeweizen:
Simplify looking for target platform upgrades by injecting them into the Maven object model and let Maven check for available upgrades. That's why this project now has a pom.xml, but is NOT part of the modules list of the parent.


romani:
@Calixte , please review

romani:
@Calixte , please review

Calixte:
> That's why this project now has a pom.xml, but is NOT part of the modules list of the parent.

@Bananeweizen I noticed that the version of this artifact (and the reference to its parent) is stuck at 10.14.2-SNAPSHOT. Should we change the release script to also bump this version?

Bananeweizen:
@Calixte Great catch. Yes, please let's bump this version with the script, too. Even though it's not part of the modules list, versioning it correctly is helpful.

PR #798, opened 2025-01-09
Use OSGi service for UI startup handler

### LINKED ISSUE #781: net.sf.eclipsecs.ui (1821)
grafik

[checkstyle.log](https://github.com/user-attachments/files/17690792/checkstyle.log)

Bananeweizen:
Bundle activators should do as little work as possible. Therefore move the registration of the window listener out of the bundle activation and instead have OSGi call back when the application has started. That's later, that improves startup performance and that should avoid race conditions on the workbench display access.

Fixes #781

### DIFF
--- a/net.sf.eclipsecs.ui/src/net/sf/eclipsecs/ui/ApplicationStartedHandler.java
+++ b/net.sf.eclipsecs.ui/src/net/sf/eclipsecs/ui/ApplicationStartedHandler.java
@@ -1,0 +1,133 @@
+//============================================================================
+//
+// Copyright (C) 2003-2023 the original author or authors.
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+//
+//============================================================================
+
+package net.sf.eclipsecs.ui;
+
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.event.EventConstants;
+import org.osgi.service.event.EventHandler;
+
+import net.sf.eclipsecs.core.jobs.AbstractCheckJob;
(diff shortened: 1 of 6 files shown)


Bananeweizen:
for reviewers: the only interesting thing is line 47 of the new event handler class. everything else is either moved code from the activator class, or generated stuff from the eclipse declarative service support.

net.sf.eclipsecs.ui/src/net/sf/eclipsecs/ui/ApplicationStartedHandler.java
```diff
@@ -0,0 +1,133 @@
...
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+//
+//============================================================================
+
+package net.sf.eclipsecs.ui;
+
+import org.osgi.service.component.annotations.Component;
```
Calixte, line 23: @Bananeweizen My Eclipse is telling me that this import cannot be resolved. Any ideas?

Bananeweizen: grafik

That package is part of the org.eclipse.osgi.services bundle, which in turn is part of our target platform. The most probable reason for you not seeing it would be that you develop against the running IDE application, not against the target platform. Enable Preferences > Plugin Development > Show current target platform in status bar, so you can always easily see what you are developing against. Then open the *.target file and hit "Set as active target platform".

Calixte: I have the correct target platform.
The only way I have found to solve this is to add `org.osgi.service.component.annotations` to the `Import-Package` section of `net.sf.eclipsecs.ui/META-INF/MANIFEST.MF`.

```
--- a/net.sf.eclipsecs.ui/META-INF/MANIFEST.MF
+++ b/net.sf.eclipsecs.ui/META-INF/MANIFEST.MF
@@ -68,6 +68,7 @@ Import-Package: org.apache.commons.lang3;version="3.12.0",
  org.eclipse.ui.texteditor,
  org.eclipse.ui.views.markers,
  org.osgi.framework,
+ org.osgi.service.component.annotations,
  org.osgi.service.event;version="[1.4.0,2.0.0)",
  org.osgi.service.prefs
 Automatic-Module-Name: net.sf.eclipsecs.ui
```

Should I commit this?

Bananeweizen: That would be okay, since it's not wrong. You could try https://github.com/checkstyle/eclipse-cs/pull/831 first, where I explicitly added another bundle to the target platform. I noticed that the relevant package is exported by multiple different bundles in different versions. And maybe that's somehow related to the error that you see: 
grafik

PR #801, opened 2025-01-13
remove "fork me on GitHub" button

Bananeweizen:
The old version required some publicly hosted resources. We could replace that with own hosted resources, but it seems okay to just remove it. Nowadays people don't need this "fork me" button for learning how to contribute like 15 years ago.


romani (approved):
Yes, GitHub is standard now, no need to do special   buttons on this

PR #805, opened 2025-01-19
configure and run OpenRewrite

Bananeweizen:
OpenRewrite is a mass refactoring engine to which I contribute since 2 years. This PR configures it and runs several refactorings. Every refactoring kind is in a separate commit for easier review. I intend to enable more of the recipes later and to submit more such changes until we reach a state where all enabled recipes can be run periodically and all changes can be submitted in one go.

@rnveach The main checkstyle project might profit from a similar configuration even more. Here in this project a lot of the recipes don't catch all occurrences, because the classes loaded by Tycho are not available to the type system inside OpenRewrite. In the main project everything is plain maven, therefore each recipe should apply everywhere. If you want me to configure something (like the "RemoveTestPrefix" recipe or similar) for a demo PR, then mention me, please.

### DIFF
--- a/net.sf.eclipsecs.checkstyle/pom.xml
+++ b/net.sf.eclipsecs.checkstyle/pom.xml
@@ -17,9 +17,11 @@
         <plugins>
             <plugin>
                 <artifactId>maven-surefire-plugin</artifactId>
+                <version>3.5.2</version>
             </plugin>
             <plugin>
                 <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.13.0</version>
             </plugin>
         </plugins>
     </build>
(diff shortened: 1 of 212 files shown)


romani:
Checkstyle is not enforced over plugin due a lot of violations.
Is it make it close to enforcing?

romani (approved):
Let's try this auto fixer.

Bananeweizen:
> Checkstyle is not enforced over plugin due a lot of violations. Is it make it close to enforcing?

No. While this change fixes some of them, there are still thousands of violations. As I mentioned already in #388 and other issues, it would be simple to fix 90% of those several thousands violations in an automated way, e.g. using OpenRewrite, the Eclipse clean actions and the Eclipse formatter. I already did that locally in 2022 within 4 hours of work. However, these fixes for the 90% of violations would lead to the patch filter raising many of the remaining 10% of issues because they happen to be in the same lines. And **I don't want to do any manual changes** for these 10% just to pass the diff filter. I just want to commit the automated changes for the 90% of issues **ignoring the diff filter**, because it fixes some thousand violations, and requiring zero remaining violations in 20.000 changed lines is just not useful.
And as long as that is rejected (as it has been every time I suggested it), I'll not even start doing the 90% fixes.

romani:
@Bananeweizen , please keep @Calixte  in loop of all changes before merge.

Calixte:
> And as long as that is rejected (as it has been every time I suggested it), I'll not even start doing the 90% fixes.

We did merge #435 at some point. I think we only have a couple sevntu-checkstyle violations left.

PR #914, opened 2025-11-13
Issue #903: update to Checkstyle 12.1.0

### LINKED ISSUE #903: upgrade to checkstyle 12.1.0
https://checkstyle.org/releasenotes.html#Release_12.1.0

vtintillier:
Fixes #903

image

### DIFF
--- a/net.sf.eclipsecs.checkstyle/.classpath
+++ b/net.sf.eclipsecs.checkstyle/.classpath
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
-	<classpathentry exported="true" kind="lib" path="checkstyle-12.0.1-all.jar" sourcepath="checkstyle-checkstyle-12.0.1.zip"/>
+	<classpathentry exported="true" kind="lib" path="checkstyle-12.1.0-all.jar" sourcepath="checkstyle-checkstyle-12.1.0.zip"/>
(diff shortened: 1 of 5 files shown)


Calixte:
Our P2 repo is massive and GitHub Pages chokes on it.
This one won't go through: https://github.com/checkstyle/eclipse-cs-update-site/actions/runs/19341048508
I will have to fix this before we can move forward.

vtintillier:
Looks like we ship a lot of dependencies in our update site since #475. Not sure it was intended to include so much...

vtintillier:
I think the intent was to add some bundles that are automatically wrapped from maven artifacts during the build, but that a quick fix was to just include all dependencies. Maybe we can just list those explicitly to be included in the p2 repo. 

Also, we already have the major version number in the p2 url when installing the plugin. Maybe we could instead have one GitHub repository per major version?

Calixte:
> Maybe we can just list those explicitly to be included in the p2 repo.

This seems to do the trick: https://github.com/checkstyle/eclipse-cs/pull/915

> Maybe we could instead have one GitHub repository per major version?

I worry this would be too much overhead. @romani Any thoughts?

romani:
>Or we could make another "legacy" repository and move them there

Yes, looks like the only good solution for now.
We can have repository with content before 10.0.0, split by some numbers.

@Calixte , do you need anything from me to make it?

PR #979, opened 2026-05-04
Issue #451: add DesignForExtension to checkstyle config

Calixte:


romani:
@Calixte , thanks a lot for activation of Checks and fixing issues, as you finish, please pin me we can add eclipse-cs to regression no-error testing in main repo, so all nes violations caused by version bump will be sent to you before checkstyle is released :). So version bump will be smooth dependabot update.

PR #1059, opened 2026-09-07
Issue #101: map LeftCurly LAMBDA to brace_position_for_lambda_body

### LINKED ISSUE #101: LeftCurly FormatterSettings missing for LAMBDA
Hi,

I use the actual checkstyle plugin (version 8.12.0.201808161509).
It seems, that the transformer to codeformatter does not support the token "LAMBDA".
No errors in Eclipse message log.
The expression will simply be ignored while creating the formatter.

In the code itsself, the curly bracket is marked as wrong formatted. So, this function works correct.

If I'm correct, the following code is missing in the eclipse-cs/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/transformer/ctransformerclasses/LeftCurlyTransformer.java (Line 77):
`      } else if (tok.equals("LAMBDA")) {
        userFormatterSetting("brace_position_for_lambda_body", option);
      }`

I'm using Eclipse Oxygen.3a (Release 4.7.3a)

AzazelSensei:
LeftCurly ignored the LAMBDA token when converting a Checkstyle config to an Eclipse formatter profile, so lambda braces never got a brace_position setting.

I map LAMBDA to brace_position_for_lambda_body (and include it in the default token list, which Checkstyle already uses).

Fixes #101

### DIFF
--- a/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/transformer/ctransformerclasses/LeftCurlyTransformer.java
+++ b/net.sf.eclipsecs.core/src/net/sf/eclipsecs/core/transformer/ctransformerclasses/LeftCurlyTransformer.java
@@ -35,7 +35,7 @@ public class LeftCurlyTransformer extends AbstractCTransformationClass {
     public FormatterConfiguration transformRule() {
         String tokens = getAttribute("tokens");
         if (tokens == null) {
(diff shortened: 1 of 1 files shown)


AzazelSensei:
Closing this after 15 days without a comment.

